PR #48, opened 2015-02-14
Translated to russian till line 44 

Wagan:
While only a modest beginning...


zoic21:
It's good begin, I merge it into last beta of jeedom

PR #57, opened 2015-02-21
Doc Update

Kschorsch:


MarieCocci:
ok. Loic macht es. / loic did it manually

PR #126, opened 2015-05-21
Improve install_dependency function

orx57:
Improvement of install_dependency function (duration of execution and readability).


orx57:
I got taken up several times, regarding OS version and typo. ;)

PR #173, opened 2015-10-31
jeedom don't have to manage system file and services

saez0pub:
1) apache is not supported by jeedom
2) giving the possibility to jeedom to write in /var/log is not a possibility
3) why trying to restart fail2ban
4) cron don't needs to be restarted by nowadays

These points throws errors during installation

### DIFF
--- a/install/consistency.php
+++ b/install/consistency.php
@@ -194,17 +194,6 @@ try {
 	$cron->setTimeout(60);
 	$cron->save();
 
-	$dynamic_apache_path = dirname(__FILE__) . '/../core/config/apache_jeedom_dynamic_rules';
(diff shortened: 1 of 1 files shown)


zoic21:
Hi,
1) I fix it in jeedom 2.0
2) I can't do anythink for that jeedom is design for jeedomboard
3) it's for fix an issue with fail2ban on jeedomboard
4) Not always, on jeedomboard/rpi sometimes it don't restart after an update

saez0pub:
ok, c'était juste pour prévenir, vous pouvez clore si vous le souhaitez, pour ma part, j'applique un patch au packaging.

PR #288, opened 2016-02-09
Hide button to submit a plugin when not in expert mode

yanngarras:


zoic21:
Make manually in beta branch

PR #336, opened 2016-04-01
Memcached::Connect doesn't exist

fredleger:
Only one of these 2 options is possible :

Memcache::Connect() : https://secure.php.net/manual/en/memcache.connect.php

or 

Memcached::addServer() https://secure.php.net/manual/en/book.memcached.php


fredleger:
Memcache is old. It's better to use Memcached

PR #539, opened 2016-12-19
booleans, strict comparison, fix bug and remove ?>

ColonelMoutarde:
With booleans, only strict comparison (with === operator) should be used to lower bug risks and to improve performances.
- fix bug line 1067


ColonelMoutarde:
` 	public function setTimeout($_timeout) {
 		if ($_timeout == '' || is_string($_timeout) || is_nan(intval($_timeout)) || $_timeout < 1) {
-			$_timeout == null;
+			$_timeout = null;
 		}
 		$this->timeout = $_timeout;
 		return $this;`
Pour moi le if ne sert à rien, je ne comprend pas l'intérèt de ce code

PR #665, opened 2017-05-23
Install: set innodb_file_per_table=1 in mysql conf

desaintmartin:
In MySQL 5.5 (current version in Debian Jessie used by jeedom), innodb_file_per_table is not enabled by default.
It allows to have a .ibd file per table, allowing easier restoration in case of corruption. Since jeedom is heavily used on devices using sdcards, plaggued by corruption problems, it should be a good idea to set it.

This commit is to be reverted as soon as jeedom recommends using a version of debian shipping with MySQL 5.7 (or, better, MariaDB?).


zoic21:
Hi,
Jeedom need to work with mysql 5.5, 5.6, 5.7 and mariadb so I prefere limit the number of change parameters by defaut.

PR #909, opened 2017-12-23
Update cmd.class.js

minscof:
add day to displayDuration


minscof:
add day to displayDuration (besides hour and minute)

PR #958, opened 2018-02-23
make the filter generic + only filter cmds if the cache is set

NebzHB:
Here is a proposal about making the cmd list in cache more generic for all plugins 

and only filter if there is actual cmds in that cache var


zoic21:
J'ai intégré les modifications a la main.

NebzHB:
merci :)

PR #962, opened 2018-03-08
Somes bugs in json API

sybri:
This patch add the json header for the global api and also have the right encoding caracters


zoic21:
Hello,
Rubis code Can note work they are syntaxe error on json encode Line.

PR #1018, opened 2018-04-21
Feature/history

apevrier:
This makes sure that grouping options are similar when using history.
Corresponding translations are included.


zoic21:
Hello,
I make merge manually, it can be found in next jeedom stable version

PR #1065, opened 2018-04-28
Update history.class.php

ngreardSF:
Correction pb "Truncated incorrect time value: '-882:55:39'" lors du job


ngreardSF:
Correction pb "Truncated incorrect time value: '-882:55:39'" lors du job

PR #1136, opened 2018-09-29
bugfix in time_between function 

minscof:


minscof:
there are 2 differents requests; the most important is the bugfix for time_between function

zoic21:
Hello,
Can you redo PR in alpha branch ?

PR #1138, opened 2018-09-30
public function getPathImgIcon()  need to support docs and doc

nherreyre:
Hi, 

Image of the object are rendered with a getPathImgIcon. 
Some plugins (IFTTT, delestage,) has a 'docs' (instead of 'doc') in the subtree directory . This function should handle this case.

(or plugins needs to be corrected)

This result in broken image.

Best regards

Nicolas.


zoic21:
Bonjour,
Peux tu refaire le PR en alpha, car la il traîne pas mal de modification et je comprends pas trop pourquoi...

PR #1259, opened 2019-03-14
Using the release branch instead of the stable one

Omegagoth:
The stable branch does not exist (anymore ?) so we should use the release one instead otherwise the page could not be found and the install script is failing.

### DIFF
--- a/install/OS_specific/Docker/init.sh
+++ b/install/OS_specific/Docker/init.sh
@@ -41,7 +41,7 @@ if [ -f /var/www/html/core/config/common.config.php ]; then
 else
 	echo 'Start jeedom installation'
 	rm -rf /root/install.sh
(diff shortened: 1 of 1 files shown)


zoic21:
You should use master branch now, it's stable version.

Omegagoth:
Sorry I'm not sure I understand, you want me to recreate a pull request where I use the master branch as reference or the script should use the master branch ?
(on peut aussi parler en francais si on arrive mieux a se comprendre ;))

Omegagoth:
Ok I just have seen that you already have modified the master branch ...

Jey2k:
Cool... I wanted to check this evening... Just begin using jeedom... Shouldn't you build again the docker image too ?

PR #1271, opened 2019-03-23
update message mode

jlayec:
Why same messages are not stored ? code commented as a proposed evolution or if refused I will add a parameter

### DIFF
--- a/core/class/message.class.php
+++ b/core/class/message.class.php
@@ -131,7 +131,7 @@ class message {
 			return;
 		}
 		if ($this->getLogicalId() == '') {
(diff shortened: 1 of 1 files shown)


KiboOst:
Imagine a scenario generating same message thousand times a day ?
If you absolutely want a new message each time, include #time# tag in the message for example.

jlayec:
OK well understood to keep one occurence but at least
--> we should update the time in the DB
--> we could add a parameter to decide if we create additional messages

Waiting for your feedback before I implement the evolutions

jlayec:
new implementation and behaviour proposed in this #1350

PR #1472, opened 2019-09-10
update required php extensions

lionelortiz:
Need zip php extension to install plugins


zoic21:
Hello,
Done in alpha branch

PR #1502, opened 2019-11-17
fix css errors

MkLHX:
Hi! i have some troubles on login screen.
When i connect with my login and pwd i have a hidden message on the right of the screen
probleme_affichage_chgt_mdp_1
So when i inspect the element through browser dev-tool i find overflow hidden is hide #register content.
probleme_affichage_chgt_mdp
And the register vertical position is bad too, fix it by change the Y translation from -80% to -118%
probleme_affichage_chgt_mdp_2

I hope it's usefull.

### DIFF
--- a/desktop/css/connection.css
+++ b/desktop/css/connection.css
@@ -75,7 +75,7 @@ body {
 	background: #fff;
 	transition: all .5s;
 	color: #303030;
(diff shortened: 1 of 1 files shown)


KiboOst:
Thanks, seems this has been fixed lately.

PR #1507, opened 2019-11-23
value start for #trigger#

minscof:
missing explanation for value other


KiboOst:
Thanks, I added it into alpha which have the updated doc

PR #1535, opened 2019-12-11
Add eqLogicViewList 

mika-nt28:


KiboOst:
Any example of implementation ? Have try to put this code but seems we need some buttons to activate it in desktop/plugin.php

PR #1560, opened 2020-01-02
Add API documentation to core classes and methods

flh:
Patches in this pull request add PHPdoc comments to internal Jeedom classes, in order to ease developer understanding of Jeedom API.


flh:
I feel a bit disappointed about the closure of thus PR without explanation.
I would be sincerely grateful if you had some time for a short explanation: should I improve my contribution? Is adding this documentation pointless because of expected changes?
Regards.

PR #1572, opened 2020-01-09
Spelling corrections

jeanrobertjs:
spelling corrections


jeanrobertjs:
@zoic21 Can you tell me where I can edit the variable behind {{Coordonnées}} in line 263 for French, pls? Is there also an spelling corrections there. Thank you.

zoic21:
For translate we use transiflex, but for now we have some issue because we have v3 and v4 in same time

jeanrobertjs:
Ok so can you edit the var {{Coordonnées}} with the value "Coordonnées" on you side, pls? Thank you :D

PR #1574, opened 2020-01-13
make config and plugin persistent accross redployment

charchess:
expose port 80 and make config and plugin persistent accross redeployment


zoic21:
Fais a la main.

PR #1619, opened 2020-02-28
Update cmd.action.slider.tmplimg.html

abarrau:
Pas de "attr('data-state')" sur $('.imgCmd#uid#') ; remplacement du $(this) par $('.cmd[data-cmd_id=#id#]').


KiboOst:
Thanks for the bugfix, will integrate it but require other fix. Mobile version needs such fix also. Maybe cmd.action.slider.tmplicon also.

PR #1621, opened 2020-03-01
Use debian:stretch to create Docker images

jmjeedom:
Jeedom needs Debian Stretch to run since 3.1.5.
https://jeedom.github.io/documentation/installation/fr_FR/index#tocAnchor-1-8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:latest
+FROM debian:stretch
 
 MAINTAINER info@jeedom.com
 


KiboOst:
PR should be on alpha branche. And Jeedom v4 (alpha, beta, rc, stable v4) runs perfectly on Buster.

jmjeedom:
Understood for PR mechanism.

However, regarding the change itself:
- using the latest tag is a bit dangerous, as any update on docker hub may bring incompatibility to your product (Debian 11 for instance);
- your documentation explicitly and officially states to use stretch as mentionned in the commit message, you should either fix the documentation to favor buster, or make sure your users use stretch.

Let me know what you think.
Thanks.

KiboOst:
- Debian 11 is still in development.
- The documentation mention Jeedom v3 and you are committing on Jeedom V4, which is anyway the installed version on new Jeedom install. But yes, Documentation is being worked.
Thanks

jmjeedom:
Everything works fine with Docker on Buster, it only requires [these little changes](https://github.com/jeedom/core/pull/1622).

PR #1622, opened 2020-03-01
Fix Dockerfile mysql-client install for Debian buster

jmjeedom:
This PR fixes the creation of the docker image for Debian buster.

[a6a7a31](https://github.com/jeedom/core/commit/a6a7a310e366db0469bd050ca1d3bf3c9553ceca) comes from a [PR that was directly merged on beta](https://github.com/jeedom/core/pull/1620) by mistake.


jmjeedom:
Fixed in a different way on beta branch since then.

PR #1627, opened 2020-03-07
Unify apt operations during installation

jmjeedom:
This commit aims at reducing the number of apt invokations to save time.
On my machine, the optimization yields:
  step_2: 6'30 before, 5' after (77%).
  step_4: 3' before, 2' after (66%).

This commit also stops trying installing libav-tools and mbrola, as apt
cannot install them. It also stops trying removing brltty and net-tools,
as they are not installed. This was inferred from the following logs:

---
Package libav-tools is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  ffmpeg

Package mbrola is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'libav-tools' has no installation candidate
E: Package 'mbrola' has no installation candidate
Reading package lists...
Building dependency tree...
Reading state information...
Package 'brltty' is not installed, so not removed
Package 'net-tools' is not installed, so not removed
---


zoic21:
Hello
I'am not ok with that if one package is missing all install failed, with old slow system if on failed install continue (some package is not critical)

jmjeedom:
Fair point.

In that case, do you know which packages are optional? I can keep the spirit of that patch while enforcing your constraints.

zoic21:
It's ok for docker but for old system we need libav-tools. This script is for all jeedom platform not juste for docker.

nmaster2042:
@zoic21 what jeedom and debian version do you have ?

jmjeedom:
@nmaster2042 jeedom 4.0.45 with the latest debian buster, and I don't have any problem. But maybe I don't encounter your use case.
@zoic21 So what should I do? Do you have further guidance or should I simply drop this patch?

nmaster2042:
@jmjeedom Same version as you.
Jeedom 4.x is supported on debien strech and up.
In my point of view compatibility with old systems can't be guaranteed. 
Upgrading these will be required.

jmjeedom:
Reopening due to mistake. Not sure why it's bringing changes that already are in origin/alpha.

@zoic21 I'm still waiting for guidance. I still believe this patch has value because it saves quite some time, especially when working with Docker.

nmaster2042:
@jmjeedom: completely agree with you for this patch.

Contrary to zoïc21, I think all packages must be installed.
Some are for plugins, other for jeedom core.

Installing jeedom with missed packages may break core or plugins installad after the initial build. This apply for docker or conventional os install.

After checking, it seems libav-tools is avaliable for jessie.
For stretch and up, it's replaced by ffmpeg.

For mbrola package, in stretch and jessie it's on the non-free repo whereas on buster it's on contrib.

Based on this, install.sh could probably be adapted to detect debian version and install appropriate repo and package. 

What do you think ?

jmjeedom:
> @jmjeedom: completely agree with you for this patch.
> 
> Contrary to zoïc21, I think all packages must be installed.
> Some are for plugins, other for jeedom core.
> 
> Installing jeedom with missed packages may break core or plugins installad after the initial build. This apply for docker or conventional os install.

I'm ok with you. It seems clunky and sloppy to ignore installation failures.

> After checking, it seems libav-tools is avaliable for jessie.
> For stretch and up, it's replaced by ffmpeg.
> 
> For mbrola package, in stretch and jessie it's on the non-free repo whereas on buster it's on contrib.
> 
> Based on this, install.sh could probably be adapted to detect debian version and install appropriate repo and package.
> 
> What do you think ?

I think we should drop jessie and focus on stretch. That's what the Jeedom team is aiming at anyway.

@zoic21 Ok with you?

zoic21:
We can not drop jessie support it's about 60% of jeedom installation... We can detect debian and make it clean but it's don't change final result...

jmjeedom:
Can you please tell me which packages are optional and which are not?

I'll rewrite my patch accordingly.

PR #1641, opened 2020-03-31
Docker 123

rpi-jeedom:
Small changes to improve the Dockerfile.

@zoic21 
- I noticed there is a new Dockerfile_full. Do you want to make the same modifications? [keeping the SQL server and all the installation steps of course].
- I did not find a good reason to use supervisord, maybe you can shed some light on this.


zoic21:
Yes we need to modify also Dokkerfile_full (it's one I use to test jeedom on fly).
For supervisord because at start I see that on tuto but maybe we can use somethink like that : CMD ["/usr/sbin/apache2ctl", "-DFOREGROUND"]

PR #1642, opened 2020-03-31
[Docker] Realign Dockerfile_full with Dockerfile

rpi-jeedom:
(I did not test it)

$ diff -u Dockerfile Dockerfile_full
--- Dockerfile	2020-03-31 10:47:36.564738042 +0200
+++ Dockerfile_full	2020-03-31 12:28:53.787640813 +0200
@@ -3,12 +3,7 @@
 MAINTAINER info@jeedom.com

 COPY install/install.sh /tmp/
-RUN sh /tmp/install.sh -s 1
-RUN sh /tmp/install.sh -s 2
-RUN sh /tmp/install.sh -s 4
-RUN sh /tmp/install.sh -s 5
-RUN sh /tmp/install.sh -s 8
-RUN sh /tmp/install.sh -s 11
+RUN sh /tmp/install.sh

 COPY install/OS_specific/Docker/init.sh /root/
 RUN chmod +x /root/init.sh

### DIFF
--- a/Dockerfile_full
+++ b/Dockerfile_full
@@ -1,17 +1,10 @@
-FROM debian:latest
+FROM debian:stretch
 
 MAINTAINER info@jeedom.com
 
-RUN apt-get update
+COPY install/install.sh /tmp/
+RUN sh /tmp/install.sh
 
-ADD install/install.sh /root/install_docker.sh
-RUN chmod +x /root/install_docker.sh
-RUN /root/install_docker.sh
-
-VOLUME /var/lib/mysql
-VOLUME /var/www/html
-EXPOSE 80
-
-ADD install/OS_specific/Docker/init.sh /root/init.sh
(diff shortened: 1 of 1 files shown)


zoic21:
Why remove  : 
VOLUME /var/lib/mysql	
VOLUME /var/www/html	
EXPOSE 80

rpi-jeedom:
So we have to choose.

1/ I never use `VOLUME` in Dockerfile and use `docker run -v` because it gives more power to the user (they only create volumes if they want, we don't create volumes behind their back).
2/ `EXPOSE` is useless and only serves as [information](https://docs.docker.com/engine/reference/builder/#expose). And I like removing useless things!

That's why I removed them. If you want to keep them in `Dockerfile_full`, I'll also add them back to `Dockerfile` to align them.

rpi-jeedom:
So, @zoic21 , what do you choose?

PR #1577, opened 2020-01-14
Core/TLS checks: remove all occurences of --no-check-certificate

pierrecdn:
Because it is just insane to disable TLS checks everywhere in this project.

People should fix their setup, working in dev environments should either imply
to store a new CA in the cacert store or to access using a plain HTTP URI.

PR #1645, opened 2020-04-03
Core/TLS checks: remove last occurences of --no-check-certificate

rpi-jeedom:
Follow up of https://github.com/jeedom/core/pull/1577.


zoic21:
Hello
For URL it's needed lot of client use is own url reposotiry un https without sign certificate so we must remove check.

rpi-jeedom:
Alright I see... Thanks for the clarification.

I am not really knowledgeable with that part of the code, but it seems this is the same codepath to retrieve official updates from Jeedom, is it correct? In that case we should maybe do something like:

`
if ($url != "jeedom.com")
  $cmd .= " --no-check-certificate";
`

That would cover the main case.

zoic21:
No it's not the same it's use only by few client, 
otherwise I would of course correct it too

rpi-jeedom:
Understood! Sorry for bothering you. :)

PR #1661, opened 2020-04-12
Clean Docker's init.sh

rpi-jeedom:
With the recent changes...

- Jeedom is already installed (```step_6```), so ```[ -f /var/www/html/core/config/common.config.php ]``` is always ```true```.
- mariadb is installed inside the container, so ```[ $(which mysqld | wc -l) -ne 0 ]``` is also always ```true```.

I'm also throwing in a cosmetic change... no need to restart services that are not started.

### DIFF
--- a/install/OS_specific/Docker/init.sh
+++ b/install/OS_specific/Docker/init.sh
@@ -2,10 +2,10 @@
 echo 'Start init'
 
 echo 'Start atd'
-service atd restart
+service atd start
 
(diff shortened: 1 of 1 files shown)


zoic21:
Le docker a été revu.

rpi-jeedom:
Hey, I'm not sure why you closed this PR. The changes are only in `init.sh` still valid. :)

PR #1668, opened 2020-04-16
Clean Docker's init.sh

rpi-jeedom:
Same as https://github.com/jeedom/core/pull/1661, not sure why you closed it @zoic21 .

With the recent changes...

- Jeedom is already installed (```step_6```), so ```[ -f /var/www/html/core/config/common.config.php ]``` is always ```true```.
- mariadb is installed inside the container, so ```[ $(which mysqld | wc -l) -ne 0 ]``` is also always ```true```.

I'm also throwing in a cosmetic change... no need to restart services that are not started.


zoic21:
I don't think it's work on mount volume. When mount volume for persistance database we need to redo this action to performe database installation

edgd1er:
Hi,

In this pull you assume that everyone will put mysql server in the container. Well, there is no need to add an external service such as a database, unless you consider this container as a vm. At least, in the previous version, installing the database is an option, with this pull, it's a requirement, and service would exit on error on a system where the database server is in another container.

another point, it's possible to run systemd in a container but it requires --privileged access to the host and the /sys/fs/cgroup volume mounted so may not be the best fit for most use cases.

The s6-overlay project provides a more docker friendly process management system using s6.
Another solution is to use supervisord which is much more simple to use than s6.

rpi-jeedom:
> I don't think it's work on mount volume. When mount volume for persistance database we need to redo this action to performe database installation

So you're talking about this comment, right?
https://github.com/jeedom/core/pull/1668/commits/8b8bc6936bea8f3193171981ce70856997a22360

So do you have a list of the use cases you want to support? If so we should make them official and work on them!

rpi-jeedom:
> Hi,
> 
> In this pull you assume that everyone will put mysql server in the container.

You're right. It goes with the decision to embed the database inside the container. That's something you may want to discuss with the official maintainers. I am merely trying to clean the Docker mess here. ;)

edgd1er:
Hi @rpi-jeedom , @zoic21 

I'm not ok with embedded mysql database in jeedom, but I understand that it's easier to handle for the support when the perimeter is well defined and not depending on external elements. Nonetheless, I will continue to use my homemade image based on codafog version.

PS: in the image, you could handle service with supervisor as services are not very well handled by the lack of D-bus. supervisord is very simple and effective.

PR #1702, opened 2020-06-18
Add global parameter "Internet State"

Caelion:
With this paramater, user can specify a binary info.
Then, all plugins can use this information to check Internet state before doing some request on the web. That avoid asking in each plugin this information or launch a request to get an error.


Caelion:
Do you need more information ?
Best regards

zoic21:
Hello
Can you provide an example of plugin need that ? I don't like add option like that. The more options there are, the more difficult the software is for users

PR #1719, opened 2020-08-12
Update fr_FR.json

eliovir:
Typo


KiboOst:
Wrong way to fix it, must change in source file (here in desktop/js/plugin.js) and on alpha branch.

Thanks anyway, has been fix in alpha ;-)

PR #1734, opened 2020-10-21
typo

noodom:


KiboOst:
Many thanks. Traduction json files are generated automatically, so I have reported your corrections in core files.
;-)

PR #1738, opened 2020-10-25
Update deutsche Übersetzung

foulek57:


KiboOst:
Hi,

Many thanks for your report. We can't just merge such modifications as json file are automatically generated, but I've store your change in traduction system.

PR #1770, opened 2020-12-22
Add -s option to sudo command

bwibwi13:
Since v10 of Debian (Buster), the behavior of the sudo command has changed (see https://wiki.debian.org/NewInBuster for details).
The result is that the upgrade of Jeedom is broken on such system because all commands requested with sudo are not foudn in $PATH.

A solution is to add the -s option to the sudo command.
From my research, adding -s solve this issue and does not impact negatively other systems.

### DIFF
--- a/core/class/system.class.php
+++ b/core/class/system.class.php
@@ -84,12 +84,12 @@ class system {
 	
 	public static function getCmdSudo() {
 		if(!class_exists('jeedom')){
-			return 'sudo ';
+			return 'sudo -s ';
 		}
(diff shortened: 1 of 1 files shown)


KiboOst:
PR reverted after users report errors on Buster.

bwibwi13:
Hello @KiboOst 

I thought I had answered your comment a few weeks ago, but it seems it was not the case.

Do you have more details or bug reports about the errors reported by some users with the -s option to sudo ?

In my case, update of Jeedom systematically fails without this option.

PR #1792, opened 2021-01-16
Allow group by hour in graph's view

weimdall:
Allow the user to group by hour graph's data shown on the view page.


KiboOst:
Thanks for the hit. Will add it in both 4.1 and 4.2. 
trad json files are auto generated, modifying them is useless.

PR #1799, opened 2021-01-27
Throw error in case of issue with cleanFileSystemRight()

bwibwi13:
cleanFileSystemRight() is called at the beginning of an upgrade when "Check rights..."
In case of issues like the sudo problem I reported (see pull request #1770 ), there is an issue but the update process is not stopped because it misses a throw command in cleanFileSystemRight().
This PR checks fro the return code of the exec() command and throws an error in case of issue.
The error is then catched by the update function to abort update.

I also took the opportunity to fix an tabulation mistake that propagated until the end of the file.


KiboOst:
Isn't this fixed yet ?

bwibwi13:
Heloo @KiboOst,
Thanks for checking this PR.
I am not facing the installation error anymore. But if I read the code properly, checkFileSystemRight() still does not throw an error if rights are not available.
In case of error, the update process is not stopped during this check and fails later on..

Issue #123, opened 2015-05-18
bug when switching from 'info' to 'action' in an equipment


diving91:
Hi,
When switching from 'Info' to 'Action' in an equipment, a new  drop down list is presented with the possible actions on that equipment.
Without saving the equipment, when going back to 'Info', the drop down list is still present whereas it should disappear.

The same is also true for a dialog box in the Options. This dialog box is called "Valeur de l'info", It also should be removed when switching to 'info' type

zoic21:
Thank, it's fix in last beta of jeedom

PR #1808, opened 2021-02-16
Viewing a log file from a plugin configuration page

thanaus:
Escaping characters in the filename allowing it to be viewed correctly from the plugin configuration page. For example, the filename cloudsyncpro.#123

### DIFF
--- a/desktop/js/plugin.js
+++ b/desktop/js/plugin.js
@@ -572,9 +572,9 @@ $('#bt_savePluginLogConfig').off('click').on('click', function() {
 $('#div_configLog').on({
   'click': function(event) {
     if ($('#md_modal').is(':visible')) {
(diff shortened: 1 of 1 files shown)


zoic21:
Hello,
I think it's better to use encodeURI function.

thanaus:
Hello and thanks for the answer.

encodeURI() escapes all characters except :
`A-Z a-z 0-9 ; , / ? : @ & = + $ - _ . ! ~ * ' ( ) #`

I think it would be better to use in this case the function encodeURIComponent().
encodeURIComponent() escapes all characters except:
`A-Z a-z 0-9 - _ . ! ~ * ' ( )`

Ok for this modification ?

zoic21:
Yes ok for encodeURIComponent

PR #1816, opened 2021-03-11
function name correction

ShodanOne:
corrected byTypeAndSearhConfiguration to byTypeAndSearchConfiguration


Salvialf:
Hello,

This has already been identified for several months however it is not possible to modify the name of a function in this way. The name of the function should be changed wherever it is used including third party plugins which we cannot do.

Thanks for the PR anyway

ShodanOne:
Ok, désolé.
Je me doute bien effectivement que c’est pas neutre de modifier ça et qu’il est difficilement imaginable de coordonner ce changement avec l’ensemble des utilisateurs.

Envoyé de mon iPhone

PR #1818, opened 2021-03-25
Add unset($component['type']);

foulek57:
Add unset($component['type']); before a2o
the 'type' is not used in a2o, but get trouble if a function is called "settype" in a plugin. (get 'true' with method_exists gettype)
For example, the jMQTT plugin uses the setType() setter, when a user change a view (with the pencil), it changes the type by eqLogic in the plugin.

This modification is a prevention in case another plugin uses the same settler.

### DIFF
--- a/core/ajax/view.ajax.php
+++ b/core/ajax/view.ajax.php
@@ -137,6 +137,7 @@ try {
 				if (!is_object($eqLogic)) {
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


KiboOst:
Agree, should be to the plugin to handle this, or at least to not generate error with core functions.
Guess we can close it.

PR #1829, opened 2021-05-13
Fix humanReadableToCmd which breaks int

Domochip:
Next to PR #1825 , the problem also occurs with integers too.

in cmd::humanReadableToCmd()
If $_input contains an integer (from a parsed json), it doesn't match any case and go to the end of this function which is an str_replace().
And Bam!, 55 becomes "55"
:-(


Domochip:
Up

PR #1852, opened 2021-09-17
Update cmd.action.slider.button.html

tibo888:
Ajout d'une variable globale dynamique  permettant d'afficher un loader selon le paramètre optionnel

### DIFF
--- a/core/template/dashboard/cmd.action.slider.button.html
+++ b/core/template/dashboard/cmd.action.slider.button.html
@@ -40,7 +40,7 @@
         clearTimeout(timerHandle#uid#)
       }
       timerHandle#uid# = setTimeout(function() {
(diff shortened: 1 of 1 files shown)


KiboOst:
Sorry but I don't get it, this global parameter isn't used later in ajax call or cmd class. Would you elaborate on the goal of this and a use case ?

tibo888:
Hello, you are righ, you cannot use it from ajax or php class. 
The goal : i used an action slider (curseur) from a virtual. I used the widget "Button" for change the design. When i click on the widget the result is very long, so i used a "Optional widget parameters" ->Name = "global", Value = "true". When i click on the wigdet the loader is visible

tibo888:
I think this will be a good idea, add global: $ .issetWidgetOptParam ('# global #', 'global') on all templates

KiboOst:
After some test I don't see such behavior. being button or default is just the same.
And I don't get how to add a param that is never used in any function would change this ?

KiboOst:
This what I tested yes ...  ^^

PR #1854, opened 2021-09-21
Do not store entries for 2FA inputs

GuiPoM:
image


KiboOst:
PR should goes only on alpha branch.
For this particular case, will merge it and report into alpha.
Thanks

PR #1875, opened 2021-10-17
Update the documentation concerning available actions for the scenario HTTP API

Paul-DS:
I can see in the [German](https://github.com/jeedom/core/blob/alpha/docs/de_DE/api_http.md#szenario), [Spanish](https://github.com/jeedom/core/blob/alpha/docs/es_ES/api_http.md#gui%C3%B3n) and [Portuguese](https://github.com/jeedom/core/blob/alpha/docs/pt_PT/api_http.md#cenas) documentation that the available actions have been also translated, but from what [I can see in the code](https://github.com/jeedom/core/blob/8c57bc2e62018e4448480feaf0bddda166813579/core/api/jeeApi.php#L142), the actions are hard-coded in English.

Should I also update the documentation for others languages ?


KiboOst:
Hi,

For all language, api parameters are in english : "start", "stop", "disable" or "enable"
I've corrected documentation, thanks ;-)

Regarding PRs on documentions and traductions, here is the doc : https://doc.jeedom.com/en_US/contribute/doc

PR #1896, opened 2021-11-06
Fix log::removeAll() and add Empty All Log

BadWolf42:
1) log::removeAll() did not remove Scenario files

2) As some daemon does not re-create files and continues to write to a phantom file descriptor, it would be nice to be able to empty all log file at once.

### DIFF
--- a/core/ajax/log.ajax.php
+++ b/core/ajax/log.ajax.php
@@ -36,6 +36,12 @@ try {
 		ajax::success();
 	}
 
+	if (init('action') == 'clearAll') {
+		unautorizedInDemo();
+		log::clearAll();
+		ajax::success();
+	}
+
 	if (init('action') == 'remove') {
 		unautorizedInDemo();
 		log::remove(init('log'));
--- a/core/class/log.class.php
+++ b/core/class/log.class.php
@@ -194,6 +194,16 @@ class log {
 		}
 		return;
 	}
+	public static function clearAll() {
+		foreach (array('', 'scenarioLog/') as $logPath) {
+			$logs = ls(self::getPathToLog($logPath), '*');
+			foreach ($logs as $log) {
+				self::clear($logPath.$log);
+			}
+		}
+		return true;
+	}
+
 
 	/**
 	 * Vide le fichier de log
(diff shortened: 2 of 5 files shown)


KiboOst:
Not sure it should remove all scenario logs. These are system logs, we don't guess it should delete all scenarios logs.

BadWolf42:
Yep not sure neither, then why go explicitly through `scenarioLog/` dir in `log::removeAll(`?
https://github.com/jeedom/core/blob/ad3faf9c2a20bb3d82d5de4999f2d7e91a04ca42/core/class/log.class.php#L218-L219

My guess is that `log::removeAll(` was designed to remove all Scenario logs [after this commit](https://github.com/jeedom/core/commit/4d369017f59e746a7ab3d0beb521145fb0d13351#diff-d8f3b50d386e2232a41a44f8a4c71c0f4f1766dd569a226cb86a492491cf6951L191-R174), 
but (silently) never did as `self::remove(` is called with 2 parameters (file, path):
https://github.com/jeedom/core/blob/ad3faf9c2a20bb3d82d5de4999f2d7e91a04ca42/core/class/log.class.php#L221

And actual `log::remove(` function always had only 1 parameter:
https://github.com/jeedom/core/blob/ad3faf9c2a20bb3d82d5de4999f2d7e91a04ca42/core/class/log.class.php#L201

&nbsp;
So, there are 2 options here: 
1. keep previous behavior and remove `foreach (array('', 'scenarioLog/') as $logPath)`, as it is never used
2. apply this PR to have the "correct/new" behavior (remove all Scenario logs with `removeAll`)

What do you think?

KiboOst:
Personally I think:
When we are in analyse/logs it should remove system logs but don’t touch scenario logs 

When we are on a scenario logs it should remove scenarios logs but not system logs

zoic21:
Hello,
I mot ok to remove log of scenario when user remove log
For deamon log in 4.2 if log ended by 'd' or 'd_1' or 'd_2', Jeedom only clear log on remove.

BadWolf42:
Ok, that's clear. 👌
I will edit my PR accordingly to your vision.

BadWolf42:
Ok, I just updated and tested this PR.
removeAll and clearAll now only work for files directly in log/ folder.

Scenario log cleaning is actually done here:
https://github.com/jeedom/core/blob/8054da0fd832ba046497bd521ce3c6cc9b7c19ca/core/ajax/scenario.ajax.php#L333-L348

PR #1897, opened 2021-11-11
Update eqLogic.ajax.php

Phpvarious:
https://community.jeedom.com/t/bug-sur-la-classe-eqlogic-ajax-php-de-jeedom/71215/5

### DIFF
--- a/core/ajax/eqLogic.ajax.php
+++ b/core/ajax/eqLogic.ajax.php
@@ -526,8 +526,8 @@ try {
 					throw new Exception($e->getMessage());
 				}
 			}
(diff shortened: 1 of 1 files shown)


KiboOst:
Good catch, but your fix will return only the last saved eqLogic.
Indeed, we never save several eqLogics from the Core, but plugin.template support it. We will find a better fix to handle this.
Thanks for the feedback!

KiboOst:
Fixed : https://github.com/jeedom/core/commit/24526573bb0df945bb25be65e090a16373b4e66d

PR #1899, opened 2021-11-13
Add Cache Viewer/Editor

BadWolf42:
Button in config:
image

Modal:
image

Add a key:
image

Key added (and searched):
image

Key edit:
image

Key delete:
image

Hopes you like it ;)
Bad

### DIFF
--- a/core/ajax/cache.ajax.php
+++ b/core/ajax/cache.ajax.php
@@ -47,7 +47,21 @@ try {
 	if (init('action') == 'stats') {
 		ajax::success(cache::stats());
 	}
-	
+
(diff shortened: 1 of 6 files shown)


KiboOst:
Not sure how this will be usefull for a lot of users, I will let the team decide on this one.
Anyway, you should have desktop/js script for modals, each modal hold its own js inside. desktop/php and desktop/js are for pages.

KiboOst (changes requested):
desktop/js/cache.js should not exist, for modal js code is inside modal php file.

PR #1904, opened 2021-11-19
Update types.php

Phpvarious:
Résous les messages PHP Notice:  Undefined index dans http.error, lorsqu'une "commande" à un Generic_type qui n'est pas référencé dans INTERNAL_CONFIG.
Exemple : plugin weather --> cmdGenericType = WEATHER_RAIN
plugin googlecast --> cmdGenericType = GENERIC

### DIFF
--- a/desktop/php/types.php
+++ b/desktop/php/types.php
@@ -90,7 +90,7 @@ function jeedom_displayGenFamily($_family, $_familyId='') {
 
 			if ($cmdGenericType != '') {
 				try {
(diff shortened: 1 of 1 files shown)


KiboOst:
The try/except should do this, and ' -> ' is a separator used at different steps that won't work with just ().

KiboOst:
Fixed: https://github.com/jeedom/core/commit/248f2917ea1eb880b83341200dea3b52dda01bce

Thanks for feedback ;-)

PR #1944, opened 2022-01-25
adding the triggeringValue() function for using in scenario

pmj4:
this function returns the triggering value of a scenario
(the existing triggerValue() function returns the current value of the trigger of a scenario)

see https://community.jeedom.com/t/bloc-code-recuperation-de-la-valeur-de-la-commande-qui-a-declenche-le-scenario/75018/16

### DIFF
--- a/core/class/cmd.class.php
+++ b/core/class/cmd.class.php
@@ -1705,7 +1705,7 @@ class cmd {
 		$events = array();
 		if (!$repeat) {
 			$this->setCache(array('value' => $value, 'valueDate' => $this->getValueDate()));
-			scenario::check($this, false, $this->getGeneric_type(), $object);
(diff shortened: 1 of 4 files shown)


KiboOst:
What's the point ? Do you have usecase / testcase ?
Such implementation would break some stuff.
If you need the value at triggering time, set a tag with triggerValue() as first scenario action and use it later, even if the value change while scenario execution.

pmj4:
Hello,
did you read this post ? the test case is inside
see https://community.jeedom.com/t/bloc-code-recuperation-de-la-valeur-de-la-commande-qui-a-declenche-le-scenario/75018/16

the solution you propose is not acceptable, as I am not sure to get the right value.

what this implementation will break ?

PR #1954, opened 2022-02-07
Add more readability to inputbox on scenario

DuchkPy:
Very often the full name of the equipment does not fit in the input field.
So having `Name` and `Value` input on 2 lines helps with readability.

### DIFF
--- a/core/template/scenario/equipement.default.html
+++ b/core/template/scenario/equipement.default.html
@@ -4,6 +4,8 @@
   <span class="input-group-btn">
     <a class="btn btn-default listEquipementInfo" data-cmd_id="#id#" data-uid="#uid#" tooltip="{{Sélectionner un équipement}}" ><i class="fas fa-list-alt"></i></a>
   </span>
+</div>
+<div class="input-group input-group-sm" style="width: 100%">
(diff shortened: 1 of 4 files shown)


KiboOst:
Not fond of making scenario taking even more spaces on screen ...

DuchkPy:
You are right, it take more vertical space.

But if we compare both following images (on a 1920px wide screen), only the second one allow to be able to correctly read what is written on `Value` field :
1. Current display :
image
2. proposed display :
image

Also, sometime it's even longer leading to be very difficult to read.

KiboOst:
It adds double the space each time you edit the scenario for every action, to something that will be used once upon a time ...

Something like this may be a lot better :

scInputEdit

DuchkPy:
Your proposal is a way better, congratulation.
If it's possible to have this feature I will be more than happy. As the field is even bigger and more confortable to put everything we need inside.

DuchkPy:
@KiboOst does this feature as been implemented ?
If yes, how to access to it ?

KiboOst:
Yes this is in alpha for some time now, just ctrl+click in the field

PR #1966, opened 2022-02-14
Optimize replacement of '=' by '==' in function evaluate()

BadWolf42:
Replace all '=' by '==' in 1 pass instead of 7
Avoid messing with '==' '===' '>=' '<=' '!=' '!==' as expected

### DIFF
--- a/core/php/utils.inc.php
+++ b/core/php/utils.inc.php
@@ -1003,13 +1003,7 @@ function evaluate($_string) {
 	} else {
 		$c = 0;
 	}
-	$expr = str_replace('==', '=', $string);
-	$expr = str_replace('=', '==', $expr);
-	$expr = str_replace('<==', '<=', $expr);
-	$expr = str_replace('>==', '>=', $expr);
-	$expr = str_replace('!==', '!=', $expr);
-	$expr = str_replace('!===', '!==', $expr);
-	$expr = str_replace('====', '===', $expr);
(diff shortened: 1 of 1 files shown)


KiboOst:
Wait 4.3 to test ;-)

KiboOst:
Done some test, doesn't seems to do same thing:

$a = $b : $a == $b : OK
$a >== $b  : $a >== $b : NOK
$a !== $b : $a !== $b : NOK

BadWolf42:
Hello @KiboOst, merci pour ton retour.

En effet, l'ancien fonctionnement n'est pas respecté pour les opérateurs `>==` et `!==`.
Pour ma défense :
`>==` n'existe pas, donc lèvera un erreur
`!==` est lui légitime et ne devrait pas être modifié (comme `===`)

Souhaties-tu que je reproduise exactement le fonctionnement actuel ?

KiboOst:
I guess it should do the same thing, specially on such critical function part.
For example, >== should not raise an error, but be corrected to function like >=

@zoic21 could you confirm such assertion ?

BadWolf42:
Hi,
This regexp has the same behavior then current code:
`$expr = preg_replace("/(?|([<>!])=(?:=|===)|([^<>!]==)={1,2}|([^<>!]=)=?)/", "$1=", $string);`
 on the following test set:
`"=", "==", "===", "====", "!=", "!==", "!===", "<=", "<==", "<=>", "<==>", ">=", ">=="`

But it's less readable and understandable then previous one, 
also it's not just a simple replacement of affectation operation `=` by test operation `==`.

PR #1986, opened 2022-03-01
do not read html tags in logs

tomitomas:
if the log msg contains html tags, then they are not displayed anymore in the "colorful" display

before : 
image

after : 
image

https://community.jeedom.com/t/logs-colorees-ne-pas-afficher-les-tags-html/80130

### DIFF
--- a/core/js/log.class.js
+++ b/core/js/log.class.js
@@ -305,6 +305,7 @@ jeedom.log.colorReplacement = {
 
 }
 jeedom.log.stringColorReplace = function(_str) {
+  _str= _str.replaceAll('<', '&lt;');
   for (var re in jeedom.log.colorReplacement) {
     _str = _str.split(re).join(jeedom.log.colorReplacement[re])
   }


Mips2648:
Hi,
I do agree this shouldn't be the case, this can cause xss exploit.
But I think we shouldn't limit to one char but instead use [htmlspecialchars()](https://www.php.net/manual/en/function.htmlspecialchars.php) server side on each line of log: `htmlspecialchars(string $string, ENT_QUOTES)`

Discussion on the topic on community: https://community.jeedom.com/t/logs-colorees-ne-pas-afficher-les-tags-html/80130/3

KiboOst:
Agree htmlspecialchars should be use for security. If anyone can do a PR, will have a look at it

KiboOst:
Integrated in alpha

tomitomas:
Fix with 8d3e108c71e1d0125f111fb82566c8b561b0c70e

PR #1987, opened 2022-03-01
Fix scenario tags in "IN 0 mins"

BadWolf42:
## Proposed change

As Identified by mic78000 in Community thread:
https://community.jeedom.com/t/les-tags-ne-passent-pas-dans-les-dans/80056/13

When using "IN 0 mins" inside a scenario, tags are not passed to the "IN" execution block.

This PR fixies `core/class/scenarioElement.class.php` to passes tags to the CLI execution of `core/php/jeeScenario.php` when "scheduled now".

## Type of change
- [ ] 3rd party lib update
- [X] Bugfix (non breaking change)
- [ ] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation

## Test check
Tested on VM in a Synology NAS, with Debian 10 and a DIY Jeedom v4.2.13 fresh install
Reproduction scenario and expected behaviors is in the [Community thread](https://community.jeedom.com/t/les-tags-ne-passent-pas-dans-les-dans/80056/13) :
- Use a IN block with 0 mins delay,
- Tags should be passed to the IN block execution as per [documentation](https://doc.jeedom.com/fr_FR/core/4.2/scenario#Les%20blocs): 
image

### DIFF
--- a/core/class/scenarioElement.class.php
+++ b/core/class/scenarioElement.class.php
@@ -210,6 +210,7 @@ class scenarioElement {
 				$cmd = __DIR__ . '/../../core/php/jeeScenario.php ';
 				$cmd .= ' scenario_id=' . $_scenario->getId();
 				$cmd .= ' scenarioElement_id=' . $this->getId();
+				$cmd .= ' tags=\'' . json_encode($_scenario->getTags()).'\'';
 				$cmd .= ' >> ' . log::getPathToLog('scenario_element_execution') . ' 2>&1 &';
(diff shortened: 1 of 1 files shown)


BadWolf42:
I confirm that tags are success fully passed to AT blocks: https://github.com/jeedom/core/blob/41bb9ffd5455805d90c972c13f2a56fedb22970a/core/class/scenarioElement.class.php#L268

Test: 
image

image

PR #1992, opened 2022-03-03
Update scenarioExpression.class.php

lefilliatre:
## Proposed change
Inprovement of variable detection
Bug fix when value contain nested variables.
+ improvements

## Type of change

- [ ] 3rd party lib update
- [x ] Bugfix (non breaking change)
- [ ] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation

## Test check
correct bug when value in variable action contains nested variable/tag at different level
ex:
` Variable(var1)+durationbetween(#comm#,1,Variable(var2),date('Y-m-d'))*60`
was working but not
` Variable(var1)+durationbetween(#comm#,1,Variable(var1),date('Y-m-d'))*60`
because var1 is in two different iteration levels

### DIFF
--- a/core/class/scenarioExpression.class.php
+++ b/core/class/scenarioExpression.class.php
@@ -1174,112 +1174,120 @@ class scenarioExpression {
 		return $_default;
 	}
 
-	public static function setTags(&$_expression, &$_scenario = null, $_quote = false, $_nbCall = 0) {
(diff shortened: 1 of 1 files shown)


lefilliatre:
Remise en forme avant nouveau PR

PR #1997, opened 2022-03-04
improve diagnostics in date convertion

lefilliatre:
## Proposed change
Add log when conversion in date fail

## Type of change

- [ ] 3rd party lib update
- [x ] Bugfix (non breaking change)
- [ ] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation

## Test check
if user enter a bad date format ex 21-02-22 the converstion by strtotime fails and the entry is replace by 1970-01-01 00:00:00
without information for the end user

### DIFF
--- a/core/class/scenarioExpression.class.php
+++ b/core/class/scenarioExpression.class.php
@@ -286,8 +286,16 @@ class scenarioExpression {
 		if (!is_object($cmd) || $cmd->getIsHistorized() == 0) {
 			return '';
 		}
-		$_startTime = date('Y-m-d H:i:s', strtotime(self::setTags($_startDate)));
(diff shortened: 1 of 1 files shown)


Salvialf:
Hello, 

For sure if a bad date format is given to a function that needs a date to work,  it will not work. 

Everything is explained in documentation and these functions are working well for a while. This will too much code to maintain just for user information in complement of existant documentation.

Thanks for your proposal.

KiboOst:
Would just add that putting 8 times exact same code block is bad practice for maintenance and would require writing a function for that.

lefilliatre:
@KiboOst you're right!
I think i should ask or suggest my modification in the forum rather than post in git!
Thanks

PR #2000, opened 2022-03-17
Create a checkPHP action to check for any PHP synthax error

NebzHB:
## Proposed change
Add a new Action workflow to check php synthax on push and PR's, that allow to spot synthax errors in php on each push or PR ! (check actions tab > checkPHP)

## Type of change
automatisation for php check

### DIFF
--- a/.github/workflows/checkPHP.yml
+++ b/.github/workflows/checkPHP.yml
@@ -1,4 +1,4 @@
-name: Build
+name: checkPHP
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


zoic21:
Hello,

It's already in place since few month (year maybe).

NebzHB:
I don't see it

zoic21:
It's manage on Jenkins, on each commit Jenkins check all PHP file and other test en slack me if there is an issue.

PR #2020, opened 2022-06-18
add event for isActive modification (with setIsActive)

noodom:
add an event to notify of isActive scenario attribute modification, for use by users scenarios or plugins


zoic21:
Hello,
Are your sure about that ? Normally on change on that jeedom already send an event on save https://github.com/jeedom/core/blob/alpha/core/class/scenario.class.php#L1060

PR #2021, opened 2022-06-18
add event when a variable is updated

noodom:
add an event to notify that a variable has been updated, for use by users scenarios or plugins

### DIFF
--- a/core/class/dataStore.class.php
+++ b/core/class/dataStore.class.php
@@ -107,6 +107,9 @@ class dataStore {
 	}
 
 	public function postSave() {
+		if ($this->getType() == 'scenario') {
+			event::add('scenario::variable', array('key' => $this->getKey(), 'value' => $this->getValue()));
+		}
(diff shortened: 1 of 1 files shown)


core/class/dataStore.class.php
```diff
@@ -107,6 +107,9 @@ public function save() {
 	}
 
 	public function postSave() {
+		if ($this->getType() == 'scenario') {
```
ColonelMoutarde, line 110: use strict comparison "==="

zoic21:
Hello,
I don't understand the purpose of this PR...

sagitaz:
Voici une demande faite sur jeemate.

Mips2648:
Hi,

Tell @pyreneum to not use a variable for that but a virtual info instead and problem solved.

Salvialf:
https://github.com/JeeMateTeam/JeeMate-Project/issues/220

PR #2045, opened 2022-08-19
CSP security : Use report-uri directive to have more visibility on violations

BadWolf42:
Hello Team,

As per my tests on 4.3.0, I stumbled on various CSP issues.
They were easy to identify using Chrome Debuging Console:
image

But could be quite cryptic for a not expert eye or even be completely silent if user/dev does not go in Console.

Instead of completely disabling CSP in `apache_unsecure`, [Content-Security-Policy-Report-Only](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only) header should be enabled to help users and developers identify CSP problems to come.

## Proposed change
- Add CSP Header "Content-Security-Policy-Report-Only" to `install/apache_unsecure`.
This header only reports CSP issue but still loads the ressource.

- Force the usage of `apache_unsecure` when migrating to 4.3.0, if CSP is not already enabled.

- Fix "Apache sécurisé" and "Apache non sécurisé" in `desktop/php/system.php`, to enable `mod_headers` before restarting apache, as `mod_headers` is only enabled with [a v4.2.7 fresh install](https://github.com/Jeedom/core/commit/6658aaddbf5413e135b26878dcf22d872edb860f#diff-65845b92f5ee07169285791a57b33dd35b5792a60a4d436d2c761324a492a2c1).
(But this should be more a v4.2 fix as this is addressed by update script for v4.3)

- Add an error log in Jeedom when a CSP violation occurs (`Content-Security-Policy` header):
image

- Add a info log in Jeedom when a CSP report occurs (`Content-Security-Policy-Report-Only` header):
image

See here the 2 log messages:
image

## Type of change
- [ ] 3rd party lib update
- [x] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation
- [x] Core security

## Documentation
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only#sample_violation_report
https://web.dev/csp/


BadWolf42:
As discussed [on Commu](https://community.jeedom.com/t/conseil-integration-maps/87372/25?u=bad), this PR now logs CSP issues in javascript error modal in Jeedom:

Violation:
image

Report only:
image

PR #2069, opened 2022-10-02
Fix/2068

kwizer15:


Mips2648:
Code was mostly following k&r coding guidelines, I supposed you changed everything to psr12; the difference is mainly on the fact that curly bracket for class & method are on new line (psr12) or not (k&r)

I don't think it's a good idea because now we will have a mix of the two depending the files...

And personally, I prefer by far the k&r style, much more readable

kwizer15:
Hi. There is no coding style in this project (or you have to explicitly add one or more tools for it). But if that's the only problem, just skip the first commit, it's done for.

PR #2106, opened 2022-10-25
Fix jeedom.cmd.save() when no id is provided

BadWolf42:
## Proposed change
Fix error in ajax callback for JS function `jeedom.cmd.save()`, when `id` is not provided in `cmd` param.

## Type of change
- [ ] 3rd party lib update
- [x] Bugfix (non breaking change)
- [ ] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation

## Test check
This JS call:
```js
jeedom.cmd.save({
  cmd: {
    eqType: "jMQTT",
    eqLogic_id: "302",
    name: "NEW-CMD",
    type: "info",
    subType: "string",
    configuration: {topic: "test", jsonPath: ""}
  }
});
```
currently correctly creates the next cmd
but raises `PHP Notice:  Undefined index: id in /var/www/html/core/ajax/cmd.ajax.php on line 273` in apache.error.

After the fix, jeedom.cmd.save() works without notice.

### DIFF
--- a/core/ajax/cmd.ajax.php
+++ b/core/ajax/cmd.ajax.php
@@ -270,7 +270,7 @@ try {
 		}
 		unautorizedInDemo();
 		$cmd_ajax = jeedom::fromHumanReadable(json_decode(init('cmd'), true));
-		$cmd = cmd::byId($cmd_ajax['id']);
+		$cmd = cmd::byId(init($cmd_ajax['id']));
 		if (!is_object($cmd)) {
 			$cmd = new cmd();
 		}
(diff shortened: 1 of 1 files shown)


BadWolf42:
The init is on `$cmd_ajax['id']`, not `$cmd_ajax`

KiboOst:
https://github.com/jeedom/core/blob/alpha/core/php/utils.inc.php#L128

KiboOst:
the 'id' key should be define in cmd passed from scratch :thinking:

BadWolf42:
Ok, I can understand that it should, but it is not enforced.
I can cancel my PR if this check is not desirable.

My use case is: add a new cmd to an existing eqLogic using Core function.

Salvialf:
Hi,

You say that i'ts not enforced but it is as you can see.

It is a cmd object that is expected, so it must have an 'id' field. In the same idea, the function is used to save changes on an existing command, the 'id' field is therefore essential.

BadWolf42:
Hi, there is only a `PHP Notice` in logs and a new cmd is created, I don't see that as enforced but feel free to reject my PR if you don't think it's needed.

PR #2120, opened 2022-10-31
docker-compose with:

Alcolo47:
docker-compose with:
- front: nginx/letsencrypt/fail2ban (== swag)
- mariadb 
- jeedom in php-fpm mode, no apache, no fail2ban, no mariadb server, no ntp

To use it, you have to configure the swag service with your domain (see swag documentation)

Jeedom modifications:
Some jeedom patches are needed, but patches are not applied directly in this commit on 
jeedom code but exists under auto_patch/patch.git.diff.
This is to avoid alteration of jeedom development.

- During dependencies installation: avoid install of apache, mariadb, fail2ban, ntp
- Hearth: don't look for apache
- Update: Apply patches to new jeedom core code
- Because I use some zigbee xiaomi/aqara h1 devices, I made an zha-device-handlers patcher

Patches are store in auto_patch/patch.git.diff. This avoid alteration of jeedom development but it's not easy to read.
You can switch patch.git.diff to workspace using `auto_patch/patch_workingdir.sh` script.
You can switch back to diff fil with `auto_patch/make_patch_from_workingdir.sh` script.

The Dockerfile can be built under any of those modes.


zoic21:
Hello,
I'am ok for this exept for  : 
- patch for zigbee, it's not the place to add this
- patch of jeedom core, we need to do that on core file directly

Alcolo47:
You're right.
We can integrate core modifications directly but my main goal was to not interfere with non-docker versions.
Adding `need_*` options is quick and dirty solution. I think that introducing a platform configuration may be better.

It seems to me that integrate those modifications in core is not pertinent. But having an "official" way to have 3-parts docker TODAY with the actual version can be useful.

zigbee: I will move this patch under a personal branch. But zigbee team doesn't integrate aqara wireless buttons ... This patch is required for every one that use `lumi.remote.b28ac1`

zoic21:
Ok but I have an issue with patch of jeedom.class.php and system.class.php. For me we need to incorporate this inside core directly not with a patch.

Implemantation inside patch seem good maybe just naming, maybe something like : 
- need_apache => has_local_apache
- need_mariadb_server => has_local_mariadb

We also need to handle restart apache for some plugin dependancy maybe and remove apache cli on system page.

docker/jeedom_swag_mariadb/docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+    image: linuxserver/swag
+
+    restart: unless-stopped
+
+    environment:
+      - PUID=${PUID}
+      - PGID=${PGID}
+      - TZ=Europe/Paris
+      - URL=<<<your.domain.com>>>
+      - SUBDOMAINS=<<<my_jeedom_sib_domain>>>
```
pifou25, line 63: <<<my_jeedom_sub_domain>>> ?

zoic21:
Hello,
With recent change on docker, this PR is it still relevant ?

PR #2146, opened 2023-01-04
🔨 Allow calculValueOffset/evaluateExpression ("Formule de calcul") on every info cmd

BadWolf42:
## Proposed change
Always display "Formule de calcul" on info cmd advanced configuration modal, as per following request on Community: 
https://community.jeedom.com/t/formule-de-calcul-dans-une-commande-info-autre/93564

Main reason for this change is to allow on the fly string value modification on an info cmd.

This could already be done by setting the info cmd subtype from string to numeric or binary, save, set the "Formule de calcul", save, then switch back to string and save. So its certainly more convenient that way.

## Type of change
- [ ] 3rd party lib update
- [ ] Core new feature
- [x] UI new functionnality
- [ ] Core documentation

## Test check
Tested in dev env, but no CI/CD check included.

## Documentation
https://community.jeedom.com/t/formule-de-calcul-dans-une-commande-info-autre/93564

### DIFF
--- a/core/class/cmd.class.php
+++ b/core/class/cmd.class.php
@@ -903,29 +903,30 @@ class cmd {
 			return $_value;
 		}
 		if ($this->getType() == 'info') {
+			if ($this->getSubType() == 'numeric') { // Handle comma instead of period in a float value
+				$_value = floatval(str_replace(',', '.', $_value));
+			}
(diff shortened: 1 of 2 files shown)


Jeandhom:
Merci Bad pour ton PR.

PR #2150, opened 2023-01-13
Fix Docker init step

pifou25:
J'ai cette erreur au lancement du container jeedom  "Désolé, Je ne peux sélectionner une 10mariadb étape pour vous !"

## Proposed change

This error occurs when I launch the official Jeedom container with mouting paths for /var/www/html and db /var/lib/mysql

`docker run -p 80:80 -v $PWD/jeedom:/var/www/html -v $PWD/mysql:/var/lib/mysql --hostname jeedom --name jeedom1 jeedom/jeedom
  `

This command make the script install/OS_specific/Docker/init.sh goes to the second part (Start jeedom installation), whereas, without these volume being mouted, we just see "Jeedom is already install"

## Type of change

- [ ] 3rd party lib update
- [ x] Bugfix (non breaking change)
- [ ] Core new feature
- [ ] UI new functionnality
- [ ] Core documentation

## Test check
I can't check with the official image until it is re-build again.

## Documentation

### DIFF
--- a/install/OS_specific/Docker/init.sh
+++ b/install/OS_specific/Docker/init.sh
@@ -27,7 +27,7 @@ else
 		sed -i "s/#USERNAME#/jeedom/g" /var/www/html/core/config/common.config.php
 		sed -i "s/#PORT#/3306/g" /var/www/html/core/config/common.config.php
 		sed -i "s/#HOST#/localhost/g" /var/www/html/core/config/common.config.php
-		/root/install.sh -s 10mariadb
+		/root/install.sh -s 10
 		/root/install.sh -s 11
 	fi
 fi


edgd1er:
few thoughts:
* why put the the database in the application ? not separating database and applications in docker is rather strange. in this image docker is used as a VM.`It is because the Docker Model suggests that it is better to run a single service within a container. If you wanted to build an application that required an Apache service and a MariaDB database, you should generate two different containers.`
* mariadb is the default database server since [debian 9](https://mariadb.com/newsroom/press-releases/mariadb-replaces-mysql-as-the-default-in-debian-9/) (2017). Reverting to mysql is a specific need that should addressed locally.
* this image does not handle PID1 and therefore should not be used as a production image: see explanations for [pid1](https://petermalmgren.com/signal-handling-docker/) and [systemd](https://medium.com/swlh/docker-and-systemd-381dfd7e4628) in docker. At best this image is just a POC. it should have at least pid1 handling, volumes, logs redirections. Other images exist trying to implement jeedom into docker in a more efficient way. Jeedom has always been thought as a system and not as an application, jeedom image are between 600 to 1Gb when mysql is 60Mb.

* as for the jeedom already installed, the init process checks is jeedom is installed or not, preventing installing over an exisiting instance. I'm afraid you are using an existing container and not a plain one. According to the init script, if the database variables are set, ie this file exists, jeedom is installed otherwise run the install procedure. If your container jumps to the already installed, then the /var/www/html/core/config/common.config.php exists, ie install procedure has been ran previously. you may remove that file in the container and restart the container to rerun the install process or run `docker stop <container_name>, docker rm <container_name>, docker run ....`. A big warning though, the official container has now volumes, meaning removing the container will also remove the data and you will loose all your settings.
https://github.com/jeedom/core/blob/b8a8392901c52f1d5ec5b64e6b79126608a95858/install/OS_specific/Docker/init.sh#L4-L10

pifou25:
Hello thanks for your comment, I'll be happy to discuss and help to improve the Docker image :) But this was just a bugfix for the current image.

- I fully agree with you,  but this image is just the easiest way for end users, especially those using NAS or box with Docker, and they don't care about the Docker phylosophy... Not sure if these kind of devices provide a simple way to import several containers with some shared parameters (something like a docker-compose sheet). Anyway, almost all well-known home automation software provide this "all-in-one" container (domoticz... home-assistant... homebridge...) and it is a requirement for beginners.
- MariaDB is the database. mysql is just an alias
- We should add Supervisor for the "PID1" issue.
- As long as I run the container with mounts (for var/www/html and lib/mysql with the command above) I can drop and recreate the container without loosing my datas.

edgd1er:
Building a proper image for jeedom is quite an adventure. It was one of the first project I build an image for. I learned a lot about what to do and what to avoid. I hope you'll envoy the journey.

PR #2164, opened 2023-02-13
typeHinting

ColonelMoutarde:
## Type of change

- [x] Code quality improvements

## Documentation

### DIFF
--- a/core/class/system.class.php
+++ b/core/class/system.class.php
@@ -83,7 +83,7 @@ class system {
 		return $return;
 	}
 
-	public static function getCmdSudo() {
+	public static function getCmdSudo(): string {
 		if (!class_exists('jeedom')) {
 			return 'sudo ';
 		}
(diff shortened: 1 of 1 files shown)


KiboOst:
Not an expert for this, but
https://www.php.net/manual/en/language.types.declarations.php
8.1.0 | Returning by reference from a void function is now deprecated.

Mips2648:
I think backward support is an issue with the type of declaration because this will be enforced during runtime

an other way to cover the need of typehinting  is to add comment before functions like:
```PHP
/**
 * @return bool
 */
 function myFunctionReturnBool() {}
```

recent editors will be able to cope with that raising warnings if an issue is detected but no risk of impact, I think it is preferable to only use this as a first step
on top this allow better generation of phpDoc ;)

ColonelMoutarde:
hi, if i see the minimum declaration version of PHP in composer.json it's php7.Mmy code is php7 compliant

KiboOst:
I'm all for this, but I think it's a bit late for 4.4.
4.4 is near feature fixed and ready for beta step, and adding such thing could break any call to any function, which would need heavy testing.

ColonelMoutarde:
up

PR #2219, opened 2023-05-12
Document HTTP API capabilities for POST requests

AntoineJT:
## Proposed change

*English & French*

- Document the way to use HTTP API using POST requests and precedence of parameters through body and url 
- Highlight the pitfall of trying to use 'php://input' instead of tags

I used to read a lot of old threads on forums like the jeedom community one and this documentation, and didn't find this information in hours.
Since I've done the research work, I think we can prevent it to be redone by other people by adding it to the documentation.

I did not include that GET requests ignores request body since that seems pretty standard to do, and it seemed like noise to me.
Let me know if you think otherwise.

## Type of change

- [ ] 3rd party lib update
- [ ] Core new feature
- [ ] UI new functionnality
- [X] Core documentation

## Test check

I used Docker "latest" jeedom image to test what I document here.
I used Postman to send requests in both form-data and urlencoded formats, with screenshots below

image
image

Which resulted in those logs
image

I did more tests than what I describe here, like mixing query parameters and body content to determine precedence, but I did not include screenshots of it since it is probably not that much interesting.

## Documentation

### DIFF
--- a/docs/en_US/api_http.md
+++ b/docs/en_US/api_http.md
@@ -14,6 +14,11 @@ This API is very easy to use by simple HTTP requests via URL.
 >
 > For all of this documentation, \#API\_KEY\# corresponds to your API key, specific to your installation. To find it, go to the "General" menu → "Configuration" → "General" tab".
 
+> **Note**
+>
+> In POST requests every query parameter can be sent through the request body in form-data or x-www-form-urlencoded format. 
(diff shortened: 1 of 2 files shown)


AntoineJT:
It works exactly the same on V4-stable branch, so tell me if you want me to do another pull request on the other branch or if you will just cherry-pick this commit when merged

PR #2292, opened 2023-09-28
add parameters for hidden the jeedom header

Hotfirenet:
## Proposed change

- [ ] 3rd party lib update
- [ ] Core new feature
- [x] UI new functionnality
- [ ] Core documentation


rocket13011:
refusal : 
the fullscreen method exists on plan and view.
please use this same method and add it to the core.

moreover, it's not viable for the JS to remove the header (message, maj) as it's in the header, the JS will emit an error.

PR #2366, opened 2024-01-04
fix docker service_mariadb function

jlayec:
docker is killed by exit function in case the start is impossible
with impact in case of restarting a jeedom container with a new docker image containing a new version of mariadb which requires `rm /var/lib/mysql/ib_logfile*`

### DIFF
--- a/install/OS_specific/Docker/init.sh
+++ b/install/OS_specific/Docker/init.sh
@@ -15,7 +15,7 @@ service_mariadb(){
     service mariadb $1
     if [ $? -ne 0 ]; then
       echo "${ROUGE}Cannot start mariadb - Cancelling${NORMAL}"
-      exit 1
+      return 1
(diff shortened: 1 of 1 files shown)


zoic21:
Hello,
Can you explain ? I don't understand the exact difference....

jlayec:
Hello

`exit 1` exit the script so terminate the container
`return 1` exit only the `service_mariadb` function so continue the script and execute `rm /var/lib/mysql/ib_logfile*` commands, then try to restart the mariadb service that in most cases will work so the container will work.

it is sometimes mandatory to delete /var/lib/mysql/ib_logfile* files at least when the version of mariadb has changed (at least for some versions)

it appears when we want to restart the container (without loosing data and restoring jeedom) with a new jeedom docker image

hope the explanation is better

PR #2692, opened 2024-06-17
Exclude node_modules folders from backup

NebzHB:
like python_venv, node_modules should be excluded from backups.

it increases the backup unnecessary.
it avoid plugins to detect the dependancies should be reinstalled automatically.
it forces users to restart dependancies manually instead of automatically : bad user experience.

### DIFF
--- a/install/backup.php
+++ b/install/backup.php
@@ -129,6 +129,7 @@ try {
 		'support',
 		'backup',
 		'script/tunnel',
+		'node_modules',
 		'.git',
 		'.gitignore',
 		'.log',


zoic21:
Hello,
It's not work like it does. It's create sometimes an empty folder and with default jeedom verification system when folder is present it's say dependance ok. After restore all dependance are ok but in reality it's just empty folder and user need to go on each plugin to relaunch dependance.

NebzHB:
So it's a fix to make in some plugins, not in core... or you affect all plugins !

zoic21:
Hello,
It's affect all plugin which use the jeedom standard

zoic21:
Fix here : https://github.com/jeedom/core/commit/4b969a98253439ac1c7c38fa855eaefa3ed3fe17

Issue #2882, opened 2024-09-06
cmd getSelectModal error


Salvialf:
An error occurs when `jeedom.cmd.getSelectModal` is called from a modal for the first time : 

![image](https://github.com/user-attachments/assets/dbe7f3c7-c675-4f5f-9b77-1922997f003c)

Salvialf:
Slept on it : event was bubbling from centralized javascript, solved by adding `_event.stopImmediatePropagation()`

PR #2883, opened 2024-09-06
check if mod_insertCmd exists

Salvialf:
Just added a simple check to prevent error #2882


Hotfirenet (approved):

Salvialf:
Slept on it : event was bubbling from centralized javascript, solved by adding `_event.stopImmediatePropagation()`

PR #2890, opened 2024-09-09
docker: use jeedom password variable name

edgd1er:
Fix the database password for docker env.

## Description

My last PR introduced a variable name change that does not belong to this project. As a result, the database password configuration may be wrongly set.
This PR restore the previous name.

Fixes #

## Types of changes

## PR checklist

- x ] I have checked there is no other PR open for the same change.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -304,13 +304,13 @@ step_9_jeedom_configuration() {
 
   if [ "${INSTALLATION_TYPE}" != "docker" ];then
     echo "DROP USER 'jeedom'@'localhost';" | mariadb -uroot > /dev/null 2>&1
-    mariadb_sql "CREATE USER 'jeedom'@'localhost' IDENTIFIED BY '${MARIADB_JEEDOM_PASSWD}';"
+    mariadb_sql "CREATE USER 'jeedom'@'localhost' IDENTIFIED BY '${DB_PASSWORD}';"
     mariadb_sql "DROP DATABASE IF EXISTS jeedom;"
     mariadb_sql "CREATE DATABASE jeedom;"
     mariadb_sql "GRANT ALL PRIVILEGES ON jeedom.* TO 'jeedom'@'localhost';"
 
     cp ${WEBSERVER_HOME}/core/config/common.config.sample.php ${WEBSERVER_HOME}/core/config/common.config.php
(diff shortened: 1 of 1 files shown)


zoic21:
Can you describe more ? What is the purpose ?

edgd1er:
The variable replaced is the one used in my image, jeedom has different names for DB vars.
the proper name is DB_PASSWORD, not MYSQL_JEEDOM_PASSWORD

current init.sh
https://github.com/jeedom/core/blob/5de9f9cd3324271c5939875d47fa3fcdbe34ceb2/install/OS_specific/Docker/init.sh#L99

zoic21:
I don't understand tou replace MARIADB_JEEDOM_PASSWD by DB_PASSWORD and you say you replace with MYSQL_JEEDOM_PASSWD in common.config.php . I'am little lost

edgd1er:
> I don't understand tou replace MARIADB_JEEDOM_PASSWD by DB_PASSWORD and you say you replace with MYSQL_JEEDOM_PASSWD in common.config.php . I'am little lost

you are right, I mixed up words and projects.
Line 314 and following use MARIADB_JEEDOM_PASSWD to setup common.config.php.
This var MARIADB_JEEDOM_PASSWD does not belong to this project and is never set. the name value for this project is DB_PASSWORD.

for the record, MYSQL_JEEDOM_PASSWD is the variable name for the jeedom's docker image I've been running for 7 years . I changed few days ago my var names to match jeedom/core's and avoid situation where my PR would push the wrong variables when I backport evolution from my project to yours. As you can see, I was not as successful as I thought.

Please note that a the moment, anyone trying to setup a jeedom would have to fix manually the file common.config.php as DB_PASSWORD would be replaced with en empty value.
Docker users would have their jeedom conf ok as this file is written at each boot.

pifou25:
Hello @edgd1er 
I think we don't care about the variable name, as it is never setted anywhere, as far as I can see... (?) 
I used it in my docker-compose.yml and I still use the old name (MYSQL_JEEDOM_PASSWD)... no need to fix it once it is running :) 
And there is no `docker-compose.yml` anymore in this repo.
The variable is setted at the begining of the script with a random default value
https://github.com/jeedom/core/blob/3fbddab2b956ed3a63e5889b1f917d4fa7bc4388/install/install.sh#L417

Anyway, it is a good idea to revert to the old name and stop changing it pls 👍

edgd1er:
As you wish, i'll close the PR.
Thanks you for helping jeedom to have a better docker image. There is a lot to do.

PR #2918, opened 2024-09-30
dashboard set margin on legend

KiboOst:
add user config margin on dashboard legend

legend


Mips2648:
Is it already fix in alpha? 
if not, shouldn't you do the PR on alpha branch?

Mips2648:
In my opinion, it should be first done in alpha and then merge back in beta and eventually in stable... but too late, it's done.

PR #2986, opened 2024-11-25
Display the placeholder if there is a placeholder in "test" button for messages

NebzHB:
more logical to display the placeholder if there is one, instead of title or message that may means nothing.


NebzHB:
Hello Loic, je peux le backporter sur 4.4 ?

PR #2990, opened 2024-12-03
cmd event default value to 0

### LINKED ISSUE #2989: cmd event default value
Set a default value (`0`) for `$cmd->event()` function. In case of use of `jeedom::evaluateExpression()` that returns an empty string for `false` for example:
https://community.jeedom.com/t/fonction-evaluate-false-retourne-vide/134661/3

Salvialf:
## Description
Set a default value (`0`) for `$cmd->event()` function

Fixes #2989

## Types of changes

### DIFF
--- a/core/class/cmd.class.php
+++ b/core/class/cmd.class.php
@@ -1856,7 +1856,7 @@ class cmd {
 		return $template;
 	}
 
-	public function event($_value, $_datetime = null, $_loop = 1) {
+	public function event($_value = 0, $_datetime = null, $_loop = 1) {
 		if ($_loop > 4 || $this->getType() != 'info') {
 			return;
 		}


Salvialf:
not sure, default value may be empty too...

Mips2648:
indeed, i wouldn't do that blindly on all events
what could be done potentially is to force 0 or 1 if binary command, so forbid empty or null value
but even if this seems to make sense, this could be a "breaking change" if some users or plugins took advantage of this "flexibility"

Salvialf:
> indeed, i wouldn't do that blindly on all events what could be done potentially is to force 0 or 1 if binary command, so forbid empty or null value but even if this seems to make sense, this could be a "breaking change" if some users or plugins took advantage of this "flexibility"

Thanks a lot for your comment.

Not tested under Debian11 yet but with Debian12 a call to `$cmd->event` with no `$_value` attribute will throw an error:
```
Too few arguments to function cmd::event(), 0 passed in /var/www/html/core/class/scenarioExpression.class.php(1988) : eval()'d code on line 3 and at least 1 expected
```

Test done with `$cmd->event(jeedom::evaluateExpression(false))`

Salvialf:
Thanks for your comment.

Did some test yesterday evening and had an error `Too few arguments to function cmd::event()` with `$cmd->event(jeedom::evaluateExpression(false));`

Tested again now with no error so it must be that i made a mistake in my tests yesterday.

You're totally right, closing issue. Thanks again

PR #3019, opened 2025-01-27
Fix eqlogic delete confirmation

Mips2648:
## Description
Today, when we delete an eqLogic we see a confirmation prompt displaying too much data. 
To my understanding, this prompt is supposed to warn user in case the eqLogic is (actually) used somewhere in Jeedom to prevent futur orphans.
The issues are:
- we see command from the eqLogic itself (in case an action command is linked to an info command of the same eqLogic)
- we see the current object owning the eqLogic
- we see commands that current eqLogic is using (mostly the case for virtual but could be the case for others as well)

example with a virtual in which I copied a eqLogic from sonos:
image

while I expect to not see any warning if I delete it (because the deletion will not bring any issue); example after this PR:
image

Of course, if I try to delete the "source" eqLogic (from sonos plugin in my example), I still see the full list of commands (the one from virtual eqLogic:
image

### Suggested changelog entry
Amélioration du message de confirmation d'un eqLogic

Fixes #

## Types of changes

## PR checklist

### DIFF
--- a/core/ajax/eqLogic.ajax.php
+++ b/core/ajax/eqLogic.ajax.php
@@ -25,8 +25,8 @@ try {
 	}
 
 	ajax::init(array('uploadImage'));
-  
-  	if (init('action') == 'uploadImage') {
+
+	if (init('action') == 'uploadImage') {
 		if (!isConnect('admin')) {
 			throw new Exception(__('401 - Accès non autorisé', __FILE__));
 		}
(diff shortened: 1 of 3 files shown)


NebzHB:
good catch ! might be interesting to publish on stable too ! (no break)

NebzHB (approved):

NebzHB:
great, could it be merged ?

PR #3052, opened 2025-04-20
Fix command select using key OR value for selected

Flobul:
Fix command select on widget using key AND value for selected value.
Give the choice to use key or value from #listValue#

## Description

When #listValue# has same kind of value for key/value (i.e.` 1|7;2|6;3|5;4|4;5|3;6|2;7|1`), using the actuel condition broke the selected value.

### Suggested changelog entry

Add option in command configuration to choose between key and value for the selected listValue

Fixes #
Fix bug introduced here https://github.com/jeedom/core/pull/2675

## Types of changes

- [X] New feature _(non-breaking change which adds functionality)_

## PR checklist

- [x] I have added tests to cover my changes.

### DIFF
--- a/core/class/cmd.class.php
+++ b/core/class/cmd.class.php
@@ -1671,7 +1671,8 @@ class cmd {
 				$coupleArray = explode('|', $element);
 				$cmdValue = $this->getCmdValue();
 				if (is_object($cmdValue) && $cmdValue->getType() == 'info') {
-					if ($cmdValue->execCmd() == $coupleArray[0] || $cmdValue->execCmd() == $coupleArray[1]) {
+					$keySeclected = intval($this->getConfiguration('listValueSelected', 0));
+					if ($cmdValue->execCmd() == $coupleArray[$keySeclected]) {
 						$listOption .= '<option value="' . $coupleArray[0] . '" selected>' . $coupleArray[1] . '</option>';
 						$foundSelect = true;
 					} else {
--- a/desktop/modal/cmd.configure.php
+++ b/desktop/modal/cmd.configure.php
@@ -122,6 +122,12 @@ $configEqDisplayType = jeedom::getConfiguration('eqLogic:displayType');
                       ?>
                     </div>
                   </div>
+                  <div class="form-group">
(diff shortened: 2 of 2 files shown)


core/class/cmd.class.php
```diff
@@ -1671,7 +1671,8 @@ public function toHtml($_version = 'dashboard', $_options = '') {
 				$coupleArray = explode('|', $element);
 				$cmdValue = $this->getCmdValue();
 				if (is_object($cmdValue) && $cmdValue->getType() == 'info') {
-					if ($cmdValue->execCmd() == $coupleArray[0] || $cmdValue->execCmd() == $coupleArray[1]) {
+					$keySeclected = intval($this->getConfiguration('listValueSelected', 0));
```
kwizer15, line 1674: $keySelected (instead of $keySeclected)

core/class/cmd.class.php
```diff
@@ -1671,7 +1671,8 @@ public function toHtml($_version = 'dashboard', $_options = '') {
...
+					if ($cmdValue->execCmd() == $coupleArray[$keySeclected]) {
```
kwizer15, line 1675: $keySelected

desktop/modal/cmd.configure.php
```diff
@@ -122,6 +122,12 @@
                       ?>
                     </div>
                   </div>
+                  <div class="form-group">
+                    <label class="col-xs-4 control-label">{{Sélection par index de valeur}}</label>
+                    <div class="col-xs-8">
+                      <input type="checkbox" class="cmdAttr" data-l1key="configuration" data-l2key="listValueSelected" />
```
kwizer15, line 128: Maybe a select input will be more appropriate.

Flobul: Sure.
Should be easier to understand, I think.

kwizer15:
On the other hand, it could potentially break the existing system. I think we should make sure that if this is not configured, we keep the old behavior.

Flobul:
For me, old system since this PR https://github.com/jeedom/core/pull/2675 is buggy.

PR #3100, opened 2025-06-24
Listener remove event ?

Salvialf:
I assume this is not tested at all but why no `$listener->removeEvent($_id)` function exists ?

While listeners can handle multiple events and adding one is straightforward, removing an event currently requires several steps, which could be simplified in my opinion.

### DIFF
--- a/core/class/listener.class.php
+++ b/core/class/listener.class.php
@@ -317,6 +317,16 @@ class listener {
 		return $this;
 	}
 
+	public function removeEvent($_id) {
+		$event = $this->getEvent();
+		if (!is_array($event)) {
+			$event = array();
+		}
+		$id = trim($_id, '#');
+		$this->setEvent(array_diff($event, ['#' . $id . '#']));
+		return $this;
+	}
+
 	/*     * **********************Getteur Setteur*************************** */
 
 	/**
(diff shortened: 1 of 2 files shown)


kwizer15:
By definition, an event cannot be removed. However, we can tell the listener to stop dispatching it.

kwizer15:
Ok I understand better. I didn't have the exact context. Thanks for your clarification.

In this case, as you explained, I find that renaming the stopListeningEvent method is more meaningful.

Phpvarious:
The command will be removed from the listener, so I agree with Salvialf.
`$listener->removeEvent($_id)` is more appropriate.

core/class/listener.class.php
```diff
@@ -317,6 +317,16 @@ public function addEvent($_id) {
 		return $this;
 	}
 
+	public function removeEvent($_id) {
+		$event = $this->getEvent();
+		if (!is_array($event)) {
+			$event = array();
+		}
+		$id = trim($_id, '#');
+		$this->setEvent(array_values(array_diff($event, ['#' . $id . '#'])));
```
kwizer15, line 326: Perhaps array_filter instead of array_diff would be more appropriate.

Hotfirenet (approved):

PR #3135, opened 2025-11-10
Add functionality to list plugin package requirements

Caelion:
Add functionality to list plugin package requirements regarding pip2 and pip3 to avoid check it manually in all plugin

## Description

Add functionality to list plugin package requirements regarding pip2 and pip3 to avoid check it manually in all plugin

### Suggested changelog entry

Page de mise à jour : Ajout d'une synthèse des paquets pip2 et pip3 utilisés par les plugins et leurs versions attendues afin de limiter les mises à jours dépassant les versions requises par les plugins.

Fixes #

## Types of changes

- [X] New feature _(non-breaking change which adds functionality)_

## PR checklist

- [X] I have added tests to cover my changes.

### DIFF
--- a/desktop/php/update.php
+++ b/desktop/php/update.php
@@ -87,13 +87,13 @@ if (strpos($logUpdate, 'END UPDATE') || count(system::ps('install/update.php', '
 					</tbody>
 				</table>
 			</div>
+
 			<div role="tabpanel" class="tab-pane" id="log" style="overflow:auto;overflow-x: hidden">
 				<legend><i class="fas fa-info-circle"></i> {{Log :}}</legend>
 				<div id="div_log">
 					<pre id="pre_updateInfo"></pre>
 				</div>
 			</div>
-
 
 			<div role="tabpanel" class="tab-pane" id="os" style="overflow:auto;overflow-x: hidden">
 				<div class="alert alert-info">{{IMPORTANT : Ne sont affichés ici que les packages Linux n’étant pas à jour. Une liste vide signifiant que votre système Linux est à jour.}}</div>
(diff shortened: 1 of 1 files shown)


Mips2648:
- comment in french to be deleted
- this feature already exists in config > last tab > check system package
- and anyway, this is useless under debian 12 and following because each plugin dependencies are installed separately so you will never have conflicts between plugins
- it is not mandatory to have a package.json to manage dependencies so putting alert box in such case will create a lot of noise

Mips2648 (changes requested):

Mips2648:
@Caelion, could you please provide feedback?
Do you agree this feature already exists, and so so we can close the PR, or did we miss something?

Caelion:
Sorry for the delay.
Since the PR, I migrate to Debian 12 :
- Python2 update button and Python3 update butten are now always greyed out
- Is it what you mean by "and anyway, this is useless under debian 12 and following because each plugin dependencies are installed separately so you will never have conflicts between plugins"
- Does it mean that now, a package can be installed several times with different version at same time ?
- Regarding the other feature here : "check system package" : I didn't know this menu. Maybe we could :
  - add a column on this menu depend on the packages.json information ?
  - add a link from update menu to check system package menu ?

If you think now we can't break, you can just close it.

Mips2648:
yes, all plugins dependencies installed by the core, so via packages.json, under debian 12, it will be installed under a VENV.
Anyway, it is required for all under deb12 but it was even possible before but with custom script (like I do actually).
It means that all plugins will have their own installation of their dependencies, possibly with different version if need be.

In "Check system package", the info already depends on packages.json and if a update is required, there is already a button to proceed.

About the button to reach that page from the "Update menu", I don't think we should do that:
- I don't think it is a good idea to multiply the number of button/way to reach that page
- it remains a sensitive page, in normal circumstance an user doesn't have to do anything on it, managing dependencies from plugin config page is enough 99%; 
I think it should remains kind of "advance config menu" where you go because you know

So I think we can close this PR

Caelion:
Thank you for your complete explanation Mips.
Ok for me.

PR #3163, opened 2025-12-18
Random backup only for cloud backup enable

zoic21:


Salvialf (commented):
`if (config::byKey('market::cloudUpload', 0) == 1)` should offer better targeting

PR #3236, opened 2026-04-01
Enhance search log filter with term support

limad:
Titre : Refactor search functionality to support multiple terms and :not filter.

## Description

The log filter input (`in_searchLogFilter`) previously supported only a single search term at a time, requiring users to retype their query to switch between logs.

This refactor allows multiple comma-separated terms to be entered simultaneously (e.g. `ai_assistant,jee_mcp`), so several logs can be highlighted at once. The existing `:not()` syntax is preserved and now works per-term, meaning it can be mixed with positive terms in the same query.

No changes to the PHP side or the rest of the JS page were required.

### Suggested changelog entry

> Log filter now supports multiple comma-separated terms and per-term `:not()` exclusion.

Fixes #

## Types of changes

- [ ] Bug fix
- [ ] Breaking change

## PR checklist

- [ ] I have checked there is no other PR open for the same change.
- [ ] I have read the contribution guidelines.

### DIFF
--- a/desktop/js/log.js
+++ b/desktop/js/log.js
@@ -37,28 +37,24 @@ if (!jeeFrontEnd.log) {
 
 //searching
 document.getElementById('in_searchLogFilter')?.addEventListener('keyup', function(event) {
-  var search = event.target.value
-  if (search == '') {
+  var raw = event.target.value
+  if (raw == '') {
     jeeP.logListButtons.seen()
     return
   }
-  var not = search.startsWith(":not(")
-  if (not) {
-    search = search.replace(':not(', '')
-  }
-  search = jeedomUtils.normTextLower(search)
+
+  // Découpe sur la virgule → liste de termes (avec support :not par terme)
+  var terms = raw.split(',').map(t => t.trim()).filter(t => t.length > 0)
+
   jeeP.logListButtons.unseen()
-  var match, text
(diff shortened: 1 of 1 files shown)


desktop/js/log.js
```diff
@@ -37,28 +37,24 @@ if (!jeeFrontEnd.log) {
...
     jeeP.logListButtons.seen()
     return
   }
-  var not = search.startsWith(":not(")
-  if (not) {
-    search = search.replace(':not(', '')
-  }
-  search = jeedomUtils.normTextLower(search)
+
+  // Découpe sur la virgule → liste de termes (avec support :not par terme)
```
Mips2648, line 46: - no comment in french
- I don't think this is relevant to let a comment here as the code speak by itself

Mips2648 (changes requested):
Hello,
Thanks for this PR!
Unfortunately it will be impossible to accept because there is an ongoing workflow migration (to use the standard gitflow) on the repo and PR shouldn't be done on alpha anymore but on develop. Documentation still needs to be adapted.
We cannot rebase the PR automatically because there are some unwanted changes in alpha (due to jenkins)

Could you please resubmit your PR on develop please?

Mips2648:
new PR: https://github.com/jeedom/core/pull/3270

PR #3245, opened 2026-04-06
feat: improve "do not remove log" feature

TiTidom-RC:
## Description

edit by @Mips2648 : the below description is not relevant anymore, see comment https://github.com/jeedom/core/pull/3245#issuecomment-4207039632

Adds a `logProtect()` plugin hook in `log::remove()` so that plugins can
declare log files that should be **cleared** (emptied) instead of **deleted**.

## Why?

The list of protected logs in `log::remove()` is currently hard-coded in the
Core (`nginx.error`, `http.error`, logs ending with `d`/`d_1`/`d_2`/`d_3`).
A plugin with a log file like `myplugin_events` or `myplugin_audit` has no way
to prevent its permanent deletion when `log::remove()` is called.

## How?

Exact same pattern as `backupExclude()` already used in `install/backup.php`:
- `plugin::listPlugin(true)` + `method_exists()` + static call
- No new mechanism, no database change, no migration

Plugins that want to protect a log file simply declare:

```php
public static function logProtect(): array {
    return ['myplugin_events', 'myplugin_audit'];
}
```

## Impact

- **Fully backward-compatible**: plugins without `logProtect()` are unaffected
- `log::removeAll()` is automatically covered (it calls `self::remove()`)
- Same performance cost as `backupExclude()` in backup.php (already accepted)

## Files changed

- `core/class/log.class.php` — +9 lines in `remove()`

### Suggested changelog entry

Les plugins peuvent désormais déclarer une méthode statique `logProtect(): array` pour protéger des fichiers de log contre la suppression. Les logs listés seront vidés (`log::clear`) au lieu d'être supprimés définitivement lors d'un appel à `log::remove()`.

Fixes #

## Types of changes

## PR checklist

### DIFF
--- a/core/class/log.class.php
+++ b/core/class/log.class.php
@@ -187,16 +187,15 @@ class log extends AbstractLogger {
 	/**
 	 * Check authorisation to emptying log file
 	 */
-	public static function authorizeClearLog($_log, $_subPath = '') {
+	public static function authorizeClearLog(string $_log, string $_subPath = ''): bool {
 		$path = self::getPathToLog($_subPath . $_log);
-		return !((strpos($_log, '.htaccess') !== false)
-			|| (!file_exists($path) || !is_file($path)));
+		return !((strpos($_log, '.htaccess') !== false) || (!file_exists($path) || !is_file($path)));
 	}
 
 	/**
 	 * Empty log file
 	 */
-	public static function clear($_log) {
+	public static function clear(string $_log) {
 		if (self::authorizeClearLog($_log)) {
 			$path = self::getPathToLog($_log);
 			com_shell::execute(system::getCmdSudo() . 'chmod 664 ' . $path . '> /dev/null 2>&1;' . system::getCmdSudo() . 'chown -R ' . system::get('www-uid') . ':' . system::get('www-gid') . ' ' . $path . ' > /dev/null 2>&1;' . system::getCmdSudo() . ' cat /dev/null > ' . $path);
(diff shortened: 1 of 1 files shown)


Mips2648:
I supposed that the actual issue is that you still have file handler open on those files and so your daemon continue to write in this "deleted" file that continue to grow?

Did you saw that discussion: https://community.jeedom.com/t/jeedom-4-5-un-manque-de-logs-apres-le-passage-du-crondaily/145005/52 
I think this is related so it might be interesting to know.

But I like the idea, 2 comments:
1. in case of "removeAll", code will get list of plugins and check for existing "logProtect" method in loop for each log; maybe we can "cache" (a simple array in the class would be enough) the final list of "files to protect" once per http request (so I'm not talking about any advanced cache but only a cache within the current instance of the class; 
I agree that won't make a huge difference, it just tickles me a bit ;-)
2. I was a bit confused at the beginning because the name of the method is "logProtect" and "if true", we call clear... which seems not a way to "protect" the log ;-)
what about renaming the method to `logClearOnly` or `logDoNotRemove` to make it more explicit on the purpose?

TiTidom-RC:
Hello, 

Thanks for this review. 

Yes it's related with this topic on community but not only : i have daemon logs which are growing fast in debug on my dev jeedom, and the daily cron remove them, and yes the daemon continue to write in the deleted file, but the most important : it disepear from jeedom interface... The main goal is to protect these custom daemon logs to only empty them and not remove them 🙂 (that was the meaning of "protect" 😋

Good idea to cache the content in an class array, i will look at how to add this.

And no problem to rename the function, what do you prefer : logClearOnly ? is it clear enough and in the code style of jeedom ?

Mips2648:
I'm fine with `logClearOnly` 
I suggest you go with that if ok for you and if someone else has a better idea before the merge it will still be possible to adapt easily

Mips2648:
As discussed, a new approach has been defined:
if these 3 conditions are true:
- plugin is active
- plugin has a daemon
- log name is not exactly equal to the pluginId which is installed (ex. if log name is "virtual", it matchs the plugin "virtual"; if log name is "zwavejsd" it doesn't exactly match pluginid "zwavejs")
then the log cannot be removed (deleted), it must be "cleared"; if one of these conditions is not true, then the log can be deleted.

It means that if the log is the base log of the php code, then it can be deleted; if it's another log, most probably the log written by a deamon, then it cannot be deleted to avoid issue with daemon that continue to write in a file that doesn't exist anymore.

With these rules, the existing test of log filename ending by "d", "d_1", "d_2" or "d_3" can be removed as well to make the code more generic

This approach has the advantage to not rely on plugin to work so we don't need to adapt (nor ask all devs) code of all plugins

TiTidom-RC:
Hello,

Following the feedback on this PR, I reworked the implementation to avoid any
plugin-side change. This turned out to be more involved than expected.

The idea was: if a log name starts with an active plugin's ID, is not exactly equal
to it, and the plugin has a daemon, clear it instead of deleting it. The daemon
condition is essential — on Linux, deleting a file held open by a running process
doesn't free the inode; the process keeps writing to a ghost file until the daemon
restarts.

The tricky part: the daemon check is on the **plugin**, not on the log file itself.
A Core log can share a prefix with an active plugin that has a daemon. For example,
if a plugin named `monitoring` had a daemon, it would match `monitoring_cloud` (a Core
log) by prefix, and the algorithm would prevent its deletion — even though no daemon
process is actually writing to it.
Pure prefix detection isn't reliable enough.

**Solution:** a hardcoded exclusion list of all known Core log names. Plugins whose
ID is in that list are skipped during the prefix check. Combined with a
length-descending sort (so `Nut_free` wins over `Nut`) and `break` on first match,
all edge cases are handled correctly.

The old `endsWith('d')` / `'d_1'` / `'d_2'` / `'d_3'` rules are removed — covered
more generically by this new logic.

Mips2648 (approved):

Salvialf (approved):

PR #3253, opened 2026-04-14
fix(core): prevent null select in scenario templates after async load

Zyg0m4t1k:
## 🐛 Bug

Scenario templates use async loading (`allOrderedByGroupObjectName`) and manipulate a `<select>` element.

If the DOM changes (modal closed, navigation, reload), the `<select>` may no longer exist when the AJAX response returns.

This causes a JS error:
- select is null
- appendChild fails

## ✅ Fix

- Re-query the DOM inside the async callback
- Add guard:

```js
if (!select) return;
```

### DIFF
--- a/core/template/scenario/remove_inat.default.html
+++ b/core/template/scenario/remove_inat.default.html
@@ -11,6 +11,7 @@
     },
     success: function (scenarios) {
       let select = document.querySelector('.expressionAttr[data-uid="#uid#"][data-l1key="options"][data-l2key="scenario_id"]')
+      if (!select) return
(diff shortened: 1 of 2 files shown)


Mips2648:
Hello,

You wrote that the fix consist in "re-query the DOM" (+ add guard) but in the code the query was already there; 
Could you clarify what you mean? or was it the IA that was a bit too optimist? 

Could you please describe how we can reproduce the issue?

Mips2648:
@Zyg0m4t1k could you please come back to us?

Mips2648:
Close stale PR

PR #3289, opened 2026-04-21
Fix: translations string concatenation

Mips2648:
## Description

* Refactored log messages in `user.class.php` to remove unnecessary translation function calls for internal logs, because the source text is in english anyway and some has concatenation in it which does not work with automatic translations
* Updated `plugin.class.php` to use `sprintf` for proper variable substitution in the translated string

### Suggested changelog entry

Corrections des traductions

Fixes warning in jeetranslate workflow
image

## Types of changes

## PR checklist

- [ ] I have read the [[La ligne directrice pour contribuer à ce projet / Contribution guidelines for this project](.github/CONTRIBUTING.md)).
- [ ] I grant the project the right to include and distribute the code under the GNU.
- [ ] I have verified that the code complies with the projects coding standards.

### DIFF
--- a/core/class/user.class.php
+++ b/core/class/user.class.php
@@ -56,51 +56,51 @@ class user {
 	public static function connect(string $_login, string $_mdp) {
 		$sMdp = (!is_sha512($_mdp)) ? sha512($_mdp) : $_mdp;
 		if (config::byKey('ldap:enable') == '1' && function_exists('ldap_connect')) {
-			log::add("connection", "info", __('LDAP Authentification', __FILE__));
+			log::add("connection", "info", 'LDAP Authentification', __FILE__);
 			$ad = ldap_connect(config::byKey('ldap:host'), config::byKey('ldap:port'));
 			if (!$ad) {
-				log::add("connection", "info", __('Connection LDAP Error', __FILE__));
+				log::add("connection", "info", 'Connection LDAP Error', __FILE__);
 				return false;
 			}
-			log::add("connection", "info", __('LDAP Connection OK', __FILE__));
+			log::add("connection", "info", 'LDAP Connection OK', __FILE__);
 			ldap_set_option($ad, LDAP_OPT_PROTOCOL_VERSION, 3);
 			ldap_set_option($ad, LDAP_OPT_REFERRALS, 0);
 			if (config::byKey('ldap:tls')) {
 				if (!ldap_start_tls($ad)) {
-					log::add("connection", "debug", __('start TLS KO', __FILE__));
+					log::add("connection", "debug", 'start TLS KO', __FILE__);
 					return false;
 				} else {
-					log::add("connection", "debug", __('start TLS OK', __FILE__));
+					log::add("connection", "debug", 'start TLS OK', __FILE__);
 				}
 			}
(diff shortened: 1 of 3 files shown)


core/class/user.class.php
```diff
@@ -56,51 +56,51 @@ public static function byId($_id) {
...
-					log::add("connection", "info", __('LDAP bind user - login/password denied', __FILE__));
+					log::add("connection", "info", 'LDAP bind user - login/password denied', __FILE__);
 					return false;
 				}
 			}
-			log::add("connection", "debug", __('LDAP Bind user - OK', __FILE__));
+			log::add("connection", "debug", 'LDAP Bind user - OK', __FILE__);
 			if (config::bykey('ldap:filter:admin') == "" && config::bykey('ldap:filter:user') == "" && config::bykey('ldap:filter:restrict') == "") {
-				log::add("connection", "warning", __('LDAP Profile Check - [WARNING] None filter was set, "', __FILE__) . $_login . __('"  authenticated as an administrator', __FILE__));
+				log::add("connection", "warning", 'LDAP Profile Check - [WARNING] None filter was set, "' . $_login . '"  authenticated as an administrator');
```
kwizer15, line 89: Why do not use sprintf here ?

core/class/user.class.php
```diff
@@ -56,51 +56,51 @@ public static function byId($_id) {
...
 				$profile = 'admin';
 			} else {
 				foreach (array('admin', 'user', 'restrict', 'none') as $profile) {
 					if ($profile == 'none') {
-						log::add("connection", "warning", __('LDAP Profile Check - [WARNING] No profile has been set for "', __FILE__) . $_login . __('" user in the LDAP', __FILE__));
+						log::add("connection", "warning", 'LDAP Profile Check - [WARNING] No profile has been set for "' . $_login . '" user in the LDAP');
```
kwizer15, line 94: Same question (see other lines too)

Salvialf (approved):

core/class/user.class.php
```diff
@@ -56,51 +56,51 @@ public static function byId($_id) {
 	public static function connect(string $_login, string $_mdp) {
 		$sMdp = (!is_sha512($_mdp)) ? sha512($_mdp) : $_mdp;
 		if (config::byKey('ldap:enable') == '1' && function_exists('ldap_connect')) {
-			log::add("connection", "info", __('LDAP Authentification', __FILE__));
+			log::add("connection", "info", 'LDAP Authentification', __FILE__);
```
Salvialf: `authentication` in english 😉

Salvialf: Is `__FILE__` meant to be used as the logicalId or is it a leftover from the previous `__()` call?

core/class/user.class.php
Mips2648: Thanks for asking, I was expecting the question!

I did it first then I decided (unilateraly) to remove translations in logs because if once someone here has to read a log in IT or PT from an user, it mights be difficult 😉

In my opinion, logs shouls be in EN or FR worse case

Mips2648: on top, source is in English so not sure it will be translated

Salvialf (approved):

PR #3299, opened 2026-04-28
Add ESLint CI workflow with baseline (mirrors PHPStan setup)

limad:
## Summary

Adds a CI ESLint check for `core/js/` and `desktop/js/` with a baseline mechanism that mirrors the existing PHPStan workflow (per @kwizer15's suggestion in #3297). Catches actual bugs (`no-const-assign`, `no-redeclare`, `no-unreachable`, etc.) without enforcing style rules like `no-var` or `prefer-const` yet.

## What's added

| File | Role |
|---|---|
| `.eslintrc.json` | Bug-only ESLint rules (no-const-assign, no-redeclare, no-dupe-keys, no-unreachable, etc.) |
| `eslint-baseline.json` | Snapshot of the 175 currently-failing `{file, ruleId, message}` tuples — line numbers excluded so the baseline survives cosmetic line shifts |
| `.github/scripts/eslint-baseline-check.js` | Runs ESLint, diffs against baseline, fails only on NEW tuples |
| `.github/workflows/eslint.yml` | Runs the check on PRs and on push to `develop`. Includes an auto-update job that regenerates the baseline and opens a PR when fixes are merged (same pattern as `update-phpstan-baseline`) |
| `package.json` + `package-lock.json` | ESLint dependency only — no runtime impact |

## Why baseline-based

Upstream `core/js/` + `desktop/js/` currently emit ~175 ESLint errors (mostly `no-redeclare` from `var X` re-declarations across nested blocks, plus a few `no-unreachable`). A non-baseline approach would either:
- fail CI permanently until everything is fixed (blocks all PRs), or
- weaken rules so much that real bugs slip through.

The baseline tolerates existing issues, prevents new ones, and welcomes fixes (when a violation disappears, the auto-update job removes it from the baseline on merge).

## Style rules deferred

`no-var` and `prefer-const` are intentionally **not** enabled in this PR — they would dominate the noise. They can be re-enabled in a follow-up once the migration from `var` to `const`/`let` (PRs #3300–#3303) lands.

## Known risks / side effects
- Does not touch any existing JS files — CI/config only.
- `package.json` adds ESLint as a dev dependency; no runtime impact.
- Baseline is file-stable (no line numbers), so cosmetic changes don't break it.

## Local usage

```bash
npm install
npm run lint           # raw ESLint output
npm run lint:check     # baseline-aware check (same as CI)
npm run lint:baseline  # regenerate baseline after fixes
```

## Types of changes
- [ ] Bug fix
- [ ] New feature
[...]

PR #3300, opened 2026-04-28
Refactor var to const/let in core JS utilities (1/4)

limad:
## Summary

First of four migration PRs splitting the `var` → `const`/`let` work from #3297 by functional domain. This batch covers **11 utility/AJAX-boilerplate files**:

```
jeedom.class.js   private.class.js   core.js
cache.class.js    queue.class.js     listener.class.js
network.class.js  security.class.js  message.class.js
dataStore.class.js  repo.class.js
```

Most of these files follow the same uniform AJAX boilerplate pattern (`paramsRequired`, `paramsSpecifics`, try/catch, `domUtils.ajax(paramsAJAX)`), so the migration is largely mechanical: `var X = ...` → `const X = ...` for declarations that are never reassigned, `let` for the rest.

## Incidental fixes (caught during migration)

- **`jeedom.class.js`** — removed the top-level `var Highcharts` declaration. With `var` at top-level it became `window.Highcharts` and was overwritten by `highstock.js` loaded later. With `const`/`let` it would shadow `window.Highcharts`, leaving the local binding `undefined` and crashing `jeedom.init()` with `TypeError: Highcharts.setOptions is undefined`. Fix: rely on `window.Highcharts` directly.
- **`private.class.js`** — renamed top-level `var init` to `_init`. With `const init` it collided with the global `function init()` declared in `core.js` (`SyntaxError: redeclaration of non-configurable global property init`).
- **`private.class.js`** — `const value` → `let value` (was reassigned by `value += ''`, would have thrown `TypeError: Assignment to constant variable`).
- **`private.class.js`** — removed unused `var param = null` (dead code).

## Test plan

- [ ] Reload any Jeedom page — verify `jeedom.init()` runs without console errors
- [ ] Open a chart that uses Highcharts (history, dashboard widget) — verify it renders
- [ ] Trigger any API call going through `jeedom.private.getParamsAJAX` — verify the result handler still works
- [ ] Verify `init()` and `_init()` both work where used

Part of the split following #3297 discussion. Stacks logically with #3299 (ESLint baseline) but does not depend on it.

### DIFF
--- a/core/js/jeedom.class.js
+++ b/core/js/jeedom.class.js
@@ -55,8 +55,8 @@ if (!isset(jeedom.cache.getConfiguration)) {
 }
 
 jeedom.changes = function() {
-  const paramsRequired =[]
-  const paramsSpecifics ={
+  const paramsRequired = []
+  const paramsSpecifics = {
     global: false,
     noDisplayError: true,
     success: function(data) {
@@ -377,8 +377,8 @@ jeedom.notify = function(_title, _text, _class_name) {
 }
 
 jeedom.getStringUsedBy = function(_params) {
-  const paramsRequired =['search']
-  const paramsSpecifics ={}
+  const paramsRequired = ['search']
+  const paramsSpecifics = {}
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
@@ -396,8 +396,8 @@ jeedom.getStringUsedBy = function(_params) {
 }
 
 jeedom.getIdUsedBy = function(_params) {
-  const paramsRequired =['search']
-  const paramsSpecifics ={}
+  const paramsRequired = ['search']
+  const paramsSpecifics = {}
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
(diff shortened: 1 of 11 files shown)


core/js/jeedom.class.js
```diff
@@ -49,26 +49,25 @@ sendVarToJS([
...
-var Highcharts
 
 if (!isset(jeedom.cache.getConfiguration)) {
   jeedom.cache.getConfiguration = null
 }
 
 jeedom.changes = function() {
-  var paramsRequired = []
-  var paramsSpecifics = {
+  const paramsRequired =[]
```
kwizer15: I noticed the space after = got stripped on empty initializations: const paramsRequired =[] instead of const paramsRequired = []. It's inconsistent with the other assignments in the diff (const params = domUtils.extend(...) keeps the space) and breaks space-infix-ops.

kwizer15 (changes requested):

limad:
Good catch @kwizer15, thanks. Fixed in 148493be — restored the space after `=` on the 68 affected lines in `jeedom.class.js`. The same artefact appeared in `cmd.class.js` (PR #3301) and is fixed there in 5e57ff92. The other migration PRs (#3302, #3303) and the desktop/js work were not affected — verified with `grep -rE "(const|let)\s+\w+\s*=[\[\{]"` after the fix.

kwizer15 (approved):

Mips2648 (approved):

Salvialf (approved):

PR #3301, opened 2026-04-28
Refactor var to const/let in equipment and commands (2/4)

limad:
## Summary

Second of four migration PRs splitting #3297 by functional domain. This batch covers **4 files handling core domain logic** (equipment, commands, objects, scenarios):

```
eqLogic.class.js   cmd.class.js
object.class.js    scenario.class.js
```

These are the most complex files in `core/js/`, with deeply nested callbacks (success/error/pre_success/forEach), so this batch warrants more careful review than the utility batch.

## Incidental fix (caught during migration)

- **`cmd.class.js` (`jeedom.cmd.execute`)** — `var eqLogic` was declared inside an `if (notify) { ... }` block but referenced from callbacks at lines 87, 107, 131, 151, 166, 172 (the `jeedom.cmd.notifyEq(eqLogic, true)` calls in `pre_success`). With `var`, function-scoped hoisting made it visible. Migrated to `const`/`let` it became block-scoped and threw `ReferenceError: eqLogic is not defined` on every dashboard command execution. Fix: hoisted the declaration out of the `if` block using a ternary so `const eqLogic = ... : null`.

## Test plan

- [ ] Execute a command from the dashboard (action button or info refresh) — verify no `ReferenceError` and the eqLogic notification (the brief icon flash) still triggers
- [ ] Open the configuration modal of any equipment — verify `eqLogic.print` and `eqLogic.save` work
- [ ] Open a scenario and run it manually — verify scenario execution
- [ ] Open an object summary on the dashboard — verify summary updates

Part of the split following #3297 discussion.

### DIFF
--- a/core/js/cmd.class.js
+++ b/core/js/cmd.class.js
@@ -62,8 +62,8 @@ jeedom.cmd.execute = function(_params) {
   if (_params.value != 'undefined' && (is_array(_params.value) || is_object(_params.value))) {
     _params.value = JSON.stringify(_params.value)
   }
-  const paramsRequired =['id']
-  const paramsSpecifics ={
+  const paramsRequired = ['id']
+  const paramsSpecifics = {
     global: false,
     pre_success: function(data) {
       if (data.state != 'ok') {
@@ -205,8 +205,8 @@ jeedom.cmd.execute = function(_params) {
 }
 
 jeedom.cmd.test = function(_params) {
-  const paramsRequired =['id']
-  const paramsSpecifics ={
+  const paramsRequired = ['id']
+  const paramsSpecifics = {
     global: false,
     success: function(result) {
       switch (result.type) {
@@ -482,8 +482,8 @@ jeedom.cmd.resetUpdateFunction = function() {
 }
 
 jeedom.cmd.getWidgetHelp = function(_params) {
-  const paramsRequired =['id', 'version']
-  const paramsSpecifics ={}
+  const paramsRequired = ['id', 'version']
+  const paramsSpecifics = {}
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
@@ -503,8 +503,8 @@ jeedom.cmd.getWidgetHelp = function(_params) {
 }
 
 jeedom.cmd.toHtml = function(_params) {
-  const paramsRequired =['id', 'version']
-  const paramsSpecifics ={}
+  const paramsRequired = ['id', 'version']
+  const paramsSpecifics = {}
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
@@ -523,8 +523,8 @@ jeedom.cmd.toHtml = function(_params) {
 }
 
 jeedom.cmd.replaceCmd = function(_params) {
-  const paramsRequired =['source_id', 'target_id']
-  const paramsSpecifics ={}
+  const paramsRequired = ['source_id', 'target_id']
+  const paramsSpecifics = {}
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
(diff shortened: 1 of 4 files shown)


limad:
FYI: applied the same formatting fix here in 5e57ff92 — restored the space after `=` on 38 affected lines in `cmd.class.js`. Spotted thanks to @kwizer15 in #3300.

kwizer15 (approved):

Mips2648 (approved):

core/js/cmd.class.js
```diff
@@ -54,22 +54,22 @@ jeedom.cmd.execute = function(_params) {
...
-  var notify = _params.notify ?? true
-  if (notify) {
-    var eqLogic = document.querySelector('.cmd[data-cmd_id="' + _params.id + '"]')?.closest('div.eqLogic-widget')
-    if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)
-  }
+  const notify = _params.notify ?? true
+  const eqLogic = notify
+    ? document.querySelector('.cmd[data-cmd_id="' + _params.id + '"]')?.closest('div.eqLogic-widget')
+    : null
+  if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)
```
Salvialf: The ternary introduced for `eqLogic` works, but it feels inconsistent with the rest of the file.
In this code, ternaries are generally used only for simple value assignments, not for wrapping more complex DOM queries or conditional logic.

Jeedom typically uses explicit `if` statements for this kind of logic.
I recommend reverting to the imperative form for consistency and readability.

Mips2648: Indeed, it makes sense
I admit that I had to read these few lines 2 or 3 times to be sure there was no issue, so it is clearly a sign that readability could be improved

limad: Done — replaced with `let eqLogic = null` + explicit `if (notify) { eqLogic = ... }`. Same logic, consistent with the imperative style of the file.

core/js/cmd.class.js
```diff
@@ -54,22 +54,23 @@ jeedom.cmd.execute = function(_params) {
...
   }
-  var notify = _params.notify ?? true
+  const notify = _params.notify ?? true
+  let eqLogic = null
   if (notify) {
-    var eqLogic = document.querySelector('.cmd[data-cmd_id="' + _params.id + '"]')?.closest('div.eqLogic-widget')
-    if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)
+    eqLogic = document.querySelector('.cmd[data-cmd_id="' + _params.id + '"]')?.closest('div.eqLogic-widget') ?? null
   }
+  if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)
```
Salvialf, line 61:
```suggestion
    eqLogic = document.querySelector('.cmd[data-cmd_id="' + _params.id + '"]')?.closest('div.eqLogic-widget') ?? null
  if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)
}
```

Salvialf (changes requested):

Salvialf (approved):

core/js/cmd.class.js
Mips2648, line 88: don't we have an issue here? eqLogic might not be define, isn't it?
above we do `if (eqLogic) jeedom.cmd.notifyEq(eqLogic, false)`
so I guess we should do `if (eqLogic) jeedom.cmd.notifyEq(eqLogic, true)` ?

but is it possible to continue without eqLogic defined?

or do we consider this is out of scope of this PR?

Salvialf: You're right to question it, but in fact `jeedom.cmd.notifyEq` already handles this case internally:

```js
jeedom.cmd.notifyEq = function(_eqlogic, _hide) {
  if (!_eqlogic) {
    return
  }
```

So calling `notifyEq(eqLogic, true)` without checking `eqLogic` is safe, the function simply returns if not found.

That said, the overall logic around `notify` / `eqLogic` could definitely be clarified and cleaned up in a dedicated PR. It's preferable to keep this one focused strictly on the  `var` → `let/const` refactor.

PR #3303, opened 2026-04-28
Refactor var to const/let in admin/system + log search multi-term (4/4)

limad:
## Summary

Fourth of four migration PRs splitting #3297 by functional domain. This batch covers **12 admin/system files** plus the **multi-term log search refactor from #3270**:

```
user.class.js     plugin.class.js   log.class.js
config.class.js   cron.class.js     backup.class.js
update.class.js   recovery.class.js report.class.js
note.class.js     interact.class.js appMobile.class.js
desktop/js/log.js  (incorporates #3270)
```

## Log search multi-term (#3270)

`desktop/js/log.js` adopts the multi-term search refactor proposed in #3270, but with `const`/`let` rather than `var`. The search input now accepts comma-separated terms with per-term `:not()` exclusion:

- `error` — single term (unchanged behavior)
- `error,warning` — match any of the terms
- `:not(debug)` — exclude single term
- `error,:not(daemon)` — match `error` OR exclude `daemon`

## No incidental fixes in this batch

All files in this batch follow the uniform AJAX boilerplate pattern, so the migration is purely mechanical (`var` → `const` for never-reassigned, `let` for the rest).

## Test plan

- [ ] Open the Logs page — verify the log list and selection works
- [ ] Search a log file using the new multi-term syntax (`error,warning`, `:not(debug)`) — verify highlighting
- [ ] Open Administration → Backup, Cron, Update, Health — verify each tab loads
- [ ] Open the User management page — verify user list and edit modal
- [ ] Open the Interactions page — type a query — verify response
- [ ] Open the Plugin marketplace — install/update flow remains functional
- [ ] Verify in mobile mode (`?v=m`) that the dashboard still renders (`appMobile.class.js`)

Part of the split following #3297 discussion. After this PR lands, all `core/js/*.js` files use `const`/`let` exclusively, and the ESLint workflow from #3299 can be tightened to enforce `no-var` and `prefer-const` permanently.

### DIFF
--- a/desktop/js/log.js
+++ b/desktop/js/log.js
@@ -24,8 +24,8 @@ if (!jeeFrontEnd.log) {
       this.logListButtons = document.querySelectorAll('#ul_object .li_log')
 
       //autoclick first log:
-      var logfile = getUrlVars('logfile')
-      var log = document.querySelector('#div_displayLogList .li_log[data-log="' + logfile + '"]')
+      const logfile = getUrlVars('logfile')
+      const log = document.querySelector('#div_displayLogList .li_log[data-log="' + logfile + '"]')
       if (log != null) {
         log.click()
       } else {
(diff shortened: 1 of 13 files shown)


kwizer15 (approved):

Mips2648 (approved):

desktop/js/log.js
```diff
@@ -37,36 +37,31 @@ if (!jeeFrontEnd.log) {
...
-    if (match) {
-      _bt.seen()
-    }
+    const text = jeedomUtils.normTextLower(_bt.textContent)
+    const match = terms.some(term => {
+      const not = term.startsWith(':not(')
+      const search = jeedomUtils.normTextLower(not ? term.slice(5, -1) : term)
+      return not ? !text.includes(search) : text.includes(search)
+    })
+    if (match) _bt.seen()
```
Salvialf: The refactor from `var` to `let/const` is fine, but the changes in the log search logic introduce new functional behavior (multi‑term search, new parsing rules, etc.).

This goes beyond a mechanical refactor and changes how the feature works.
For clarity, reviewability, and to avoid mixing concerns, this functional change should be moved to a dedicated PR.

Could you extract the log search modifications into a separate PR so we can review them independently?

Mips2648: I agree
I hesitated to ask the same

(another) reason to ask for that is to think about the improve of search more globally and not only in logs

limad: You're right, that was a mistake to mix both in the same PR. I've reverted `desktop/js/log.js` in this PR to a pure mechanical `var → const/let` migration (preserving the existing single-term :not() logic). The multi-term feature has been extracted into a dedicated PR: #3313.

limad: Agreed, #3313 is intentionally kept minimal (just the multi-term comma-separated OR logic for the log list filter) so it's a clean, reviewable base. If there's appetite for a broader global search improvement across pages, it could build on top of that or be a separate discussion.

Salvialf (approved):

Mips2648 (approved):

PR #3305, opened 2026-04-28
Fix timeline folder field visibility on scenario page load

### LINKED ISSUE #3304: Timeline folder field not displayed on scenario page load when timeline option is enabled
### Description

When opening a scenario with the Timeline option already enabled, the `timeline::folder` field does not appear on page load.
It only becomes visible after manually interacting with the Timeline checkbox.

Reference: https://community.jeedom.com/t/pb-daffichage-dossier-timeline-dans-les-scenarios/148774

### Type de matériel / Hardware type

Autre / Other

### Version de Jeedom / Jeedom Version

4.5.3

### Navigateur / Browser

_No response_

### Étapes pour reproduire / Steps to reproduce

1. Open a scenario with `timeline::enable` already checked
2. The `timeline::folder` field is not visible
3. Toggle the checkbox → the field appears correctly

### Logs

_No response_

Salvialf:
The visibility of the `timeline::folder` field was previously controlled through a `click` event listener on the `#generaltab` container.
Because the checkbox state is restored without triggering a click, the visibility logic was not applied on page load.

The listener has been switched to a `change` event so the visibility is correctly updated both:

- when the scenario loads with the Timeline option already enabled
- when the user toggles the checkbox manually

This ensures consistent and expected behavior.

### DIFF
--- a/desktop/js/scenario.js
+++ b/desktop/js/scenario.js
@@ -1633,7 +1633,7 @@ function filterOptions() {
     })
 }
 
-input.addEventListener('input', filterOptions) 
+input.addEventListener('input', filterOptions)
 
 select.addEventListener('change', function(event) {
   document.querySelectorAll('.addElementTypeDescription').unseen()
@@ -1999,11 +1999,11 @@ document.getElementById('div_editScenario').querySelector('div.floatingbar').add
       _target.setAttribute('data-state', '1')
       //open code blocks for later search:
       document.querySelectorAll('#div_scenarioElement div.elementCODE.elementCollapse').forEach(_code => {
-          _code.classList.remove('elementCollapse');
-          const textarea = _code.querySelector('textarea[data-l1key="expression"]');
-          if (textarea) {
-            textarea.style.display = 'block';
-          }
+        _code.classList.remove('elementCollapse')
+        const textarea = _code.querySelector('textarea[data-l1key="expression"]')
+        if (textarea) {
+          textarea.style.display = 'block'
+        }
       })
       jeeP.setEditors()
       document.querySelectorAll('textarea[data-l1key="expression"]').unseen()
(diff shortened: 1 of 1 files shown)


zoic21 (approved):

desktop/js/scenario.js
```diff
@@ -2149,6 +2141,18 @@ document.getElementById('generaltab').addEventListener('click', function(event)
...
+  var _target = null
+  if (_target = event.target.closest('.scenarioAttr[data-l2key="timeline::enable"]')) {
+    if (_target.checked) {
+      document.querySelector('.scenarioAttr[data-l2key="timeline::folder"]').seen()
+    } else {
+      document.querySelector('.scenarioAttr[data-l2key="timeline::folder"]').unseen()
+    }
+    return
+  }
+})
```
kwizer15:
```suggestion
document.getElementById('generaltab').addEventListener('change', (event) => {
  if (event.target.closest('.scenarioAttr[data-l2key="timeline::enable"]')) {
    const folder = document.querySelector('.scenarioAttr[data-l2key="timeline::folder"]')
    event.target.checked ? folder.seen() : folder.unseen()
  }
})
```

kwizer15 (changes requested):

desktop/js/scenario.js
Salvialf: Hi @kwizer15,

Thanks for the suggestion, it actually pushed me to look deeper at this code rather than just moving it around.

I went with a direct listener on the checkbox instead of keeping delegation on `#generaltab`. Since this element is static in the DOM there's no benefit to delegation here, it just adds unnecessary overhead by catching every `change` event bubbling up through the tab.

As for the `const` + ternary style, I prefer keeping it consistent with the surrounding `if/else` blocks in the file. Besides, since it's an `if/else`, only one `querySelector` is ever executed at runtime anyway, the same as storing it in a `const`.

kwizer15 (approved):

Mips2648 (approved):

PR #3313, opened 2026-05-05
feat(log): multi-term comma-separated search with :not() exclusion

limad:
## Summary

Enhances the log filter search to support multiple comma-separated terms, each optionally negated with `:not()`.

### Before
Single-term search: `error` or `:not(daemon)`

### After
Multi-term, comma-separated:
- `error,warning` → show entries matching **error** OR **warning**
- `error,:not(daemon)` → show entries matching **error** OR **not** matching **daemon**
- `:not(debug),:not(info)` → hide debug and info entries

### Logic
Each comma-separated term is evaluated independently against the log name. An entry is shown if **any** term matches (OR semantics). Each term can be negated individually.

### Known risks / side effects
- Only `desktop/js/log.js` is modified — no PHP changes, no impact on other pages.
- The existing single `:not()` syntax continues to work as before (backward compatible).
- No plugin impact (filter is UI-only, client-side).

## Notes
- Initially proposed in #3270 (closed, superseded by this PR)
- Extracted from #3303 at request of @Salvialf and @Mips2648 to keep functional changes separate from the `var→const/let` refactoring
- As noted by @Mips2648, this opens the door to a more global search improvement across pages

## Types of changes
- [ ] Bug fix
- [x] New feature *(non-breaking change which adds functionality)*
- [ ] Breaking change

## PR checklist
- [x] I have read the contribution guidelines.

## Testing
1. Open Logs page
2. Type `error,warning` → only logs containing "error" or "warning" shown
3. Type `:not(debug)` → logs NOT containing "debug" shown
4. Type `http,:not(daemon)` → logs containing "http" OR not containing "daemon" shown
5. Clear filter → all logs shown again

### DIFF
--- a/desktop/js/log.js
+++ b/desktop/js/log.js
@@ -37,28 +37,26 @@ if (!jeeFrontEnd.log) {
 
 //searching
 document.getElementById('in_searchLogFilter')?.addEventListener('keyup', function(event) {
-  var search = event.target.value
-  if (search == '') {
+  var raw = event.target.value
+  if (raw == '') {
     jeeP.logListButtons.seen()
     return
   }
-  var not = search.startsWith(":not(")
-  if (not) {
-    search = search.replace(':not(', '')
-  }
-  search = jeedomUtils.normTextLower(search)
+
+  // Multi-term: comma-separated terms, each term can be prefixed with :not()
+  // e.g. "error,warning" → show entries matching either
+  //      "error,:not(daemon)" → show entries matching "error" but not "daemon"
+  var terms = raw.split(',').map(function(t) { return t.trim() }).filter(function(t) { return t.length > 0 })
+
   jeeP.logListButtons.unseen()
-  var match, text
   jeeP.logListButtons.forEach(_bt => {
-    match = false
-    text = jeedomUtils.normTextLower(_bt.textContent)
-    if (text.includes(search)) {
-      match = true
-    }
-    if (not) match = !match
-    if (match) {
-      _bt.seen()
-    }
+    var text = jeedomUtils.normTextLower(_bt.textContent)
+    var match = terms.some(function(term) {
+      var not = term.startsWith(':not(')
+      var search = jeedomUtils.normTextLower(not ? term.slice(5, -1) : term)
+      return not ? !text.includes(search) : text.includes(search)
+    })
+    if (match) _bt.seen()
   })
 })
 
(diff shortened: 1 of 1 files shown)


limad:
Label suggestion for maintainers: `changelog-feat` + milestone `4.6` (non-breaking UI feature, one file only).

Mips2648:
It is not yet planned because there are still at least 2 unresolved questions that need alignement with the team
- validate the search format: comma or sapce or ...
- identify others search box where it would make sense to have such feature to make sure to keep coherence in the UI

On top, we need to adapt the documentation as well, or to to define how to explain the feature in the UI directly.

Mips2648 (changes requested):
On top, this PR is a regression regarding var declaration

limad:
Thanks @Mips2648 — `var` regression fixed, all 6 declarations in the new handler are now `const`.

Understood on the broader points: happy to wait for team alignment on the separator format and the scope across other search boxes before this moves forward.

desktop/js/log.js
```diff
@@ -37,27 +37,26 @@ if (!jeeFrontEnd.log) {
...
-  let search = event.target.value
-  if (search == '') {
+  const raw = event.target.value
+  if (raw == '') {
     jeeP.logListButtons.seen()
     return
   }
-  const not = search.startsWith(":not(")
-  if (not) {
-    search = search.replace(':not(', '')
```
zoic21, line 47: I think we need to handle also ), currently if we do :not(toto), search will search toto) and not toto

limad:
An idea to move this “comma or space” debate forward — and where should we discuss it, here on Community ?

desktop/js/log.js
limad:
```
document.getElementById('in_searchLogFilter')?.addEventListener('keyup', function(event) {
  const raw = event.target.value
  if (raw == '') {
    jeeP.logListButtons.seen()
    return
  }

  const terms = raw.split(',').map(t => t.trim()).filter(t => t.length > 0)
  const positive = terms.filter(t => !t.startsWith(':not('))
  const negative = terms.filter(t => t.startsWith(':not(')).map(t => jeedomUtils.normTextLower(t.slice(5, -1)))

  jeeP.logListButtons.unseen()
  jeeP.logListButtons.forEach(_bt => {
    const text = jeedomUtils.normTextLower(_bt.textContent)
    //const text = jeedomUtils.normTextLower(_bt.getAttribute('data-log'))
    const passNeg = negative.every(s => !text.includes(s))
    const passPos = positive.length === 0 || positive.some(t => text.includes(jeedomUtils.normTextLower(t)))
    if (passNeg && passPos) _bt.seen()
  })
})
```
Don’t know if this could help or if it’s the intended use ?
Working examples: 
- error, warning
- :not(ai_assistant),:not(alexaapiv2)
- error,:not(api)

PR #3319, opened 2026-05-07
fix(desktop/js): add try/catch guards for JSON.parse and empty catch blocks (scenario, widgets, massedit, recovery, plan3d)

limad:
## Summary

This PR adds robustness guards for `JSON.parse` calls without try/catch and replaces empty catch blocks with `console.error`. All of these bugs pre-existed the `var→const/let` migration. This PR contains **no style or refactor changes**.

### scenario.js
- `JSON.parse(localStorage.getItem('jeedomScCopy'))` in the paste handler had no error handling → any corrupted or manually edited localStorage entry would crash the paste action with an unhandled `SyntaxError`. Wrapped in try/catch → `showAlert` on parse failure.
- Keydown global handler (`Ctrl+S`, `Ctrl+L`, `Ctrl+Z`) had no guard for "not on scenario page" → DOM lookups inside would throw `TypeError` when triggered from other pages. Added guard: `if (!document.querySelector('.scenarioAttr[data-l1key=id]')) return`
- Two empty `catch (err) { }` blocks around context menu initialization silently swallowed initialization errors → replaced with `console.error('scenario context menu init failed', err)`

### widgets.js
- Both JSON upload handlers called `JSON.parse(e.target.result)` with no error handling → a malformed JSON file would crash the `onload` callback, leaving the UI in a loading state. Wrapped in try/catch → `showAlert('Fichier json invalide.')`.

### massedit.js
- `JSON.parse(e.target.result)` was called **before** the `try { ... }` block that was meant to protect it → a malformed import file would throw before the catch could handle it. Moved the parse inside a dedicated try/catch.

### recovery.js
- `JSON.parse(_data)` in the progress polling callback had no error handling → a corrupt API response would silently break the recovery progress UI. Wrapped in try/catch → `console.error`.

### plan3d.js
- Empty `catch (error) { }` in the window resize handler silently swallowed THREE.js renderer errors → replaced with `console.error('plan3d resize failed', error)`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Suggested changelog entry

```
- fix(desktop/js): add try/catch guards for JSON.parse calls and replace empty catch blocks (scenario, widgets, massedit, recovery, plan3d)
```

## Test plan

| File | Action | Expected |
|---|---|---|
| `scenario.js` | Corrupt `jeedomScCopy` in localStorage, click paste button | Alert shown, no unhandled exception |
| `scenario.js` | Press Ctrl+S from a non-scenario page (e.g. dashboard) | No JS error in console |
| `scenario.js` | Open scenario page, right-click on an element | Context menu appears; any init error logged to console instead of silently ignored |
| `widgets.js` | Upload a malformed JSON file as widget template | Alert "Fichier json invalide." shown, UI not frozen |
| `massedit.js` | Upload a malformed JSON file as massedit config | Alert "Fichier json invalide." shown |
| `recovery.js` | Simulate corrupt API response for progress endpoint | `console.error` logged, no crash |
| `plan3d.js` | Resize window while plan3d is open | Any renderer error logged to console instead of silently ignored |

## Known risks

- `scenario.js` keydown guard: `.scenarioAttr[data-l1key=id]` is the canonical marker that the scenario editor DOM is loaded. If this selector ever changes in a future refactor, the guard would stop working (Ctrl+S etc. would silently no-op everywhere). Low risk for now.
- All other changes are mechanical wrapping of existing calls — no behavioral change on the happy path.

### DIFF
--- a/desktop/js/massedit.js
+++ b/desktop/js/massedit.js
@@ -241,7 +241,13 @@ if (!jeeFrontEnd.massedit) {
         var readFile = new FileReader()
         readFile.readAsText(_fileHdl)
         readFile.onload = function(e) {
-          var massEditData = JSON.parse(e.target.result)
+          var massEditData
+          try {
+            massEditData = JSON.parse(e.target.result)
+          } catch (err) {
+            jeedomUtils.showAlert({ message: '{{Fichier json invalide.}}', level: 'danger' })
+            return
+          }
           var newFilter
           try {
             //type:
--- a/desktop/js/plan3d.js
+++ b/desktop/js/plan3d.js
@@ -331,6 +331,7 @@ document.body.registerEvent('cmd::update', function(_event) {
 })
 
 window.registerEvent('resize', function() {
+  if (!jeeFrontEnd.plan3d.camera || !jeeFrontEnd.plan3d.renderer) return
   jeeFrontEnd.plan3d.SCREEN_WIDTH = document.getElementById('div_display3d').offsetWidth
   jeeFrontEnd.plan3d.SCREEN_HEIGHT = document.getElementById('div_display3d').offsetHeight - 5
   try {
--- a/desktop/js/plan3d.js
+++ b/desktop/js/plan3d.js
@@ -337,7 +337,7 @@ window.registerEvent('resize', function() {
     jeeFrontEnd.plan3d.camera.aspect = jeeFrontEnd.plan3d.SCREEN_WIDTH / jeeFrontEnd.plan3d.SCREEN_HEIGHT
     jeeFrontEnd.plan3d.camera.updateProjectionMatrix()
     jeeFrontEnd.plan3d.renderer.setSize(jeeFrontEnd.plan3d.SCREEN_WIDTH, jeeFrontEnd.plan3d.SCREEN_HEIGHT)
-  } catch (error) { }
+  } catch (error) { console.error('plan3d resize failed', error) }
 }, false)
 
 window.registerEvent('dblclick', function(event) {
--- a/desktop/js/recovery.js
+++ b/desktop/js/recovery.js
@@ -81,7 +81,11 @@ if (!jeeFrontEnd.recovery) {
         async: false,
         success: function(_data) {
           if (_data) {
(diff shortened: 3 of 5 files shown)


Mips2648:
Adding `console.err` or `jeedomUtils.showAlert` isn't really bugfixing something...
What we miss in this PR is the purpose; what do you try to achieve?

limad:
> Adding `console.err` or `jeedomUtils.showAlert` isn't really bugfixing something... What we miss in this PR is the purpose; what do you try to achieve?

Thanks for the feedback.

The goal of this PR is to fix cases where invalid external data causes an unhandled JS exception that silently breaks a user action — not just to add logging.

Specifically:

widgets.js / massedit.js — invalid JSON import: before, a malformed file caused an uncaught exception inside the FileReader callback; the UI gave no feedback and the import silently failed. Now the user gets a controlled error message.
massedit.js specifically: JSON.parse was called before the existing try/catch, so the catch block was never actually protecting the import.
scenario.js — clipboard paste: a corrupted localStorage value could crash the paste handler entirely; now the action is cleanly aborted.
recovery.js: an invalid JSON response from the server would break the progress polling loop mid-recovery.
plan3d.js — early return: the resize handler accessed camera.aspect when no 3D scene was loaded, causing a TypeError on any browser resize before a plan3D is opened.
I understand that replacing catch {} with console.error alone is not a functional bugfix. If you prefer, I can split this PR: keep only the changes that fix observable user-facing behaviour (the JSON.parse guards, the plan3d early return, the scenario clipboard guard), and move the empty-catch logging replacements out to a separate PR or drop them entirely.

Mips2648:
Thanks for the additional context.
At the moment we’re not going to add defensive try/catch blocks and extra alerts/logging for these cases. We don’t have a reliable way to reproduce the reported issues, and we also want to avoid showing overly technical/debug-style messages to end users for situations that are not clearly actionable.

If you can provide a concrete reproduction path (exact page, steps, and an example of the data/input that triggers the exception), we can revisit and consider a targeted fix with user-friendly handling where it makes sense.

PR #3324, opened 2026-05-07
fix(desktop/js): correct logic bugs in scenario undo stack, massedit jValue guard, user event handler (Lot 3)

limad:
## Context

Third batch of logic bug fixes in `desktop/js/`. Each fix addresses a silent incorrect behavior (no JS exception thrown, wrong result silently returned).

### scenario.js — undo stack dedup never fires
```js
// Before (bug): array reference comparison is always false between two distinct arrays
if (newStack == this.undoStack[state - 1]) return

// After: compare by content
if (JSON.stringify(newStack) === JSON.stringify(this.undoStack[state - 1])) return
```
The early-return guard meant to avoid duplicate undo states never triggered, causing unbounded stack growth on every scenario change.

### massedit.js — jValue always captured regardless of disabled state
```js
// Before (bug): .disabled is a boolean — `!= null` is always true
if (edit.querySelector('.inputEditJValue').disabled != null) {

// After:
if (!edit.querySelector('.inputEditJValue').disabled) {
```
The jValue field was read and sent even when its input was disabled (not applicable), overwriting the correct value.

### user.js — event handler falls through after #bt_supportAccess
```js
// Before (bug): missing return + brace at column 0
    }
}

// After:
    }
    return
  }
```
Without `return`, the delegated click handler continued evaluating subsequent `if (_target = closest(...))` conditions after handling the support access button, potentially triggering unintended actions.

## Test
- Locally applied and validated: no console errors on scenario, massedit, user pages
- scenario: undo stack dedup now works correctly on repeated identical states
- massedit: jValue only captured when input is enabled
- user: support access click no longer falls through to other handlers


Mips2648:
No PR on alpha
Please check if issues still exist on develop and re-submit only what is really a bug

PR #3325, opened 2026-05-08
fix(desktop/js): null guards for dynamic querySelector and missing returns in event handlers (Lot 4)

limad:
## Context

Fourth batch of logic bug fixes in `desktop/js/`, found by static analysis. All issues are silent failures: no exception in normal conditions, but incorrect or missing behaviour.

### object.js — wrong variable name in dashboard cleanup
`removeId` is undefined in the success callback scope; the correct variable is `id` (declared 10 lines above). The `bt_gotoDashboard` menu links were never cleaned after deleting an object.

### overview.js — querySelector before .closest() without null guard
`updateSummary` is called from a MutationObserver. The added node's className may not yet be present as a DOM element. Fixed with early return.

### types.js — querySelector without null guard (×7 occurrences in 2 places)
- Generic type modal builder: `eqLogic[data-id]` and `cmd[data-id]` lookups can return null if the equipment/command is no longer in the filtered list. Added `if (!el) continue` guards.
- Context menu callback: all 5 accesses to `cmd[data-id]` replaced with a single lookup + early return `if (!cmdEl) return`.

### scenario.js — missing return in 3 event handlers
- `.scenario_link` click: fell through to `.action_link` handler, opening a command configuration dialog unexpectedly.
- `#bt_cancelElementSave` / `#bt_crossElementSave`: fell through to the `expressionAttr` Ctrl+click handler, triggering an unintended prompt.

### backup.js — querySelector on dynamic repo name
Both `.bt_restoreRepoBackup[data-repo]` and `.sel_restoreCloudBackup[data-repo]` use a value from the API that may not match any rendered element. Added null guards.

### replace.js — querySelector on dynamic eqlogicId
`resetEqlogicSelect` uses a parameter that may not match any rendered eqLogic row. Added null guard.

### view.js — querySelector before scrollIntoView
`data-zone_id` is read from a clicked element; the corresponding zone may not be rendered in the current view. Added null guard.

### administration.js — querySelector before .innerHTML in search
The tab anchor lookup can return null if a tabpanel has no matching `<a>` link. The existing `if (tabName != null)` check ran after the crash. Added early return before the access.

## Test
Locally applied. No console errors on affected pages (scenario, types, overview, replace, view, administration, backup).


Mips2648:
No PR on alpha
Please check if issues still exist on develop and re-submit only what is really a bug

PR #3328, opened 2026-05-08
fix(jeedom.class.js): guard catch blocks against undefined jeedom.private

limad:
## Problem

All 34 catch blocks in `core/js/jeedom.class.js` use the pattern:

```js
(_params?.error || paramsSpecifics.error || jeedom.private.default_params.error)(e)
```

If `jeedom.private` is undefined (e.g. `private.class.js` fails to load), the catch block itself throws a **second uncaught TypeError**:
> `Cannot read properties of undefined (reading 'default_params')`

This is especially critical for the **16 functions with `paramsRequired = []`** (`jeedom.version`, `jeedom.changes`, `jeedom.backup`, etc.): their `try` block can **only** throw if `jeedom.private` is itself undefined — meaning their catch block is guaranteed to fail in that scenario, swallowing the real error and emitting a confusing one instead.

Reported at: `core/js/jeedom.class.js:963` — `jeedom.version` catch block.

## Fix

Use optional chaining to make the catch block resilient:

```js
(_params?.error || paramsSpecifics.error || jeedom.private?.default_params?.error || console.error)(e)
```

- `jeedom.private?.default_params?.error` — safely returns `undefined` if `jeedom.private` is not yet defined
- `|| console.error` — fallback so the original error is never silently lost
- `_params?.error` — defensive guard in case the function is called without arguments

Applied to all **34 catch blocks** (1 in `jeedom.changes`, 33 in other API functions).

## Testing

- Dashboard, widgets (`jeedom.version()` call), backup, administration pages tested
- No `Uncaught TypeError` in console on any page
- Normal AJAX flows unaffected (optional chaining only activates on the `undefined` path)

### DIFF
--- a/core/js/jeedom.class.js
+++ b/core/js/jeedom.class.js
@@ -118,7 +118,7 @@ jeedom.changes = function() {
   try {
     jeedom.private.checkParamsRequired(paramsRequired)
   } catch (e) {
-    (paramsSpecifics.error || jeedom.private.default_params.error)(e)
+    (paramsSpecifics.error || jeedom.private?.default_params?.error || console.error)(e)
     return
   }
   const params = domUtils.extend({}, jeedom.private.default_params, paramsSpecifics)
@@ -382,7 +382,7 @@ jeedom.getStringUsedBy = function(_params) {
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
-    (_params.error || paramsSpecifics.error || jeedom.private.default_params.error)(e)
+    (_params?.error || paramsSpecifics.error || jeedom.private?.default_params?.error || console.error)(e)
     return
   }
   const params = domUtils.extend({}, jeedom.private.default_params, paramsSpecifics, _params || {})
@@ -401,7 +401,7 @@ jeedom.getIdUsedBy = function(_params) {
   try {
     jeedom.private.checkParamsRequired(_params || {}, paramsRequired)
   } catch (e) {
-    (_params.error || paramsSpecifics.error || jeedom.private.default_params.error)(e)
+    (_params?.error || paramsSpecifics.error || jeedom.private?.default_params?.error || console.error)(e)
     return
   }
   const params = domUtils.extend({}, jeedom.private.default_params, paramsSpecifics, _params || {})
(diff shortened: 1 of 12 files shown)


Mips2648:
Thank you for the contribution.
AI can be used as an assistance tool, but we do not accept pull requests that were identified and written entirely by AI. Submitting changes that were only discovered and produced by AI is not useful for this project.
Because of that, we cannot review or merge this PR.
Please submit a PR based on your own analysis and implementation instead.

limad:
"All the bugs reported in my latest PRs were discovered after automated audits targeting every potentially affected file (following the 'refactor(desktop/js)...' PR).

What matters is not the origin or nature of a PR, but its consistency and usefulness for the user community—though that might just be my personal opinion.

There are many subjects we've been discussing for years that have never been addressed, likely due to a lack of time. Let's take advantage of AI to move faster and better."

Salvialf:
We understand your point, but we don’t plan to refactor parts of the core based on hypothetical issues raised by automated audits or AI tools.

If a real, reproducible malfunction is identified, we’ll fix it.
But we won’t spend time rewriting working code for theoretical edge cases that don’t occur in actual Jeedom usage.

limad:
"I experienced the bug in question ('Cannot read properties of undefined (reading 'default_params')') myself this afternoon.

As I am running the 'develop' branch on my local setup, this bug—which I had never encountered before—might have been a side effect of my previous PRs. I took the opportunity to use AI to debug it and fix it at the source.

Naturally, I understand your perspective as well. However, I believe it’s better to refactor now to anticipate future bugs and focus on improving features and design... there is so much to do! We shouldn't let the Jeedom solution, which we value so much, grow outdated."

Salvialf:
Thanks for the details.

Since you mention that this error appeared only after your previous PRs were merged, and that you had never seen it before, it strongly suggests that the issue was introduced by those changes rather than coming from the core itself.

I’m running the develop branch updated daily and I haven’t encountered this error at any point.

Without logs, context, or a reproducible case on a clean develop installation, we can’t justify applying additional modifications to the core.

If the issue was introduced by one of your earlier PRs, it should be fixed directly at that level rather than by refactoring unrelated core logic. We don’t want to introduce new potential issues by changing the core without a clearly identified and reproducible problem.

PR #3341, opened 2026-05-11
Fix & improvement: add safe guard in scenario in "for" loop, "wait" & "sleep" actions

Mips2648:
## Description

As mentioned in comments [here](https://github.com/jeedom/core/issues/3164#issuecomment-3876220442) and [here](https://github.com/jeedom/core/issues/3164#issuecomment-3925571298) we should prevent infinite loop and implement safeguard in case a scenario is executed with invalide configuration.

During development of [the "while block"](https://github.com/jeedom/core/issues/3164), I realized that it is not the case for the FOR, WAIT and SLEEP that could run for almost infinite time so I propose to fix this

I took the opportunity to restructure the code and remove duplicated code.

### Suggested changelog entry
- Sécurisation des scénarios:
  - limitation du temps d’exécution de la boucle **for** à maximum 1h 
  - limitation de l'action **sleep** à maximum 1h
  - limitation de l'action **wait** à maximum 3600 occurrences (~1h). La valeur par défaut est de 3600
 - Corrections: les tags sont dorénavant correctement pris en compte lors de l'évaluation des limites pour les actions **sleep** & **wait**

Fixes #

## Types of changes

## PR checklist

### DIFF
--- a/core/class/scenarioElement.class.php
+++ b/core/class/scenarioElement.class.php
@@ -202,9 +202,15 @@ class scenarioElement {
 			if (!is_numeric($limits)) {
 				throw new Exception(__('La condition pour une boucle doit être numérique :', __FILE__) . ' ' . $limits);
 			}
+			$endTime = time() + 3600;
 			$return = false;
 			for ($i = 1; $i <= $limits; $i++) {
 				$return = $this->getSubElement('do')->execute($_scenario);
+				if (time() > $endTime) {
+					$_scenario->setLog(__('[For] Arrêt de la boucle pour cause de durée d\'exécution trop longue', __FILE__));
+					return false;
+				}
+				sleep(1);
 			}
 			return $return;
 		} elseif ($this->getType() == 'while') {
@@ -232,7 +238,7 @@ class scenarioElement {
 				$return = $this->getSubElement('do')->execute($_scenario);
 				if (time() > $endTime) {
 					$_scenario->setLog(__('[While] Arrêt de la boucle pour cause de durée d\'exécution trop longue', __FILE__));
-					return;
+					return false;
 				}
 				sleep(1);
 				$result = $this->getSubElement('while')->execute($_scenario);
--- a/core/class/scenarioElement.class.php
+++ b/core/class/scenarioElement.class.php
@@ -202,7 +202,7 @@ class scenarioElement {
 			if (!is_numeric($limits)) {
 				throw new Exception(__('La condition pour une boucle doit être numérique :', __FILE__) . ' ' . $limits);
 			}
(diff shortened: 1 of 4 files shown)


Salvialf:
**General note for reviewers**

After this PR, the three constructs will have different upper bounds:
- `for` and `while` ~1h, 
- `wait` ~2h (7200 occurrences).

Since `wait`'s 7200 limit was not properly enforced before this PR (tags not evaluated, duplicated logic), this is actually the right moment to align all three to a consistent maximum. Capping `wait` at 3600 occurrences as well would give a uniform 1h ceiling across all loop/wait constructs without introducing a real breaking change.

@zoic21, @Sekiro-kost ?

core/class/scenarioElement.class.php
```diff
@@ -202,9 +202,15 @@ public function execute(&$_scenario = null) {
...
 			}
+			$endTime = time() + 3600;
 			$return = false;
 			for ($i = 1; $i <= $limits; $i++) {
 				$return = $this->getSubElement('do')->execute($_scenario);
+				if (time() > $endTime) {
+					$_scenario->setLog(__('[For] Arrêt de la boucle pour cause de durée d\'exécution trop longue', __FILE__));
+					return false;
+				}
+				sleep(1);
```
Mips2648: Indeed, my first idea was to cap the number of iterations, but if there’s a 10‑second sleep inside the loop, it could already run for up to 10 hours (30s sleep => 30h!) — which is clearly abnormal. That’s why I opted for a time‑based limit instead, independent of the iteration count.

Salvialf (approved):

Mips2648:
I fully agree. Harmonizing their upper bounds gives us a more uniform and coherent behavior across loops and wait.

I will wait for others feedback and adapt the code accordingly

Sekiro-kost (approved):

zoic21 (approved):

Salvialf (changes requested):
Hi @Mips2648 ,

as discussed together this morning with @zoic21, we agreed to harmonize the maximum duration for all loop‑related actions (`for`, `while`, `wait`) to a single limit of 3600 seconds handled through a dedicated config variable.

We’ll wait for these changes before proceeding with the merge.

Mips2648:
> Hi @Mips2648 ,
> 
> as discussed together this morning with @zoic21, we agreed to harmonize the maximum duration for all loop‑related actions (`for`, `while`, `wait`) to a single limit of 3600 seconds handled through a dedicated config variable.
> 
> We’ll wait for these changes before proceeding with the merge.

commits done

Salvialf (approved):

PR #3342, opened 2026-05-11
refactor(dom.ui): seen(), unseen() and toggle() use hidden class consistently

Salvialf:
## Summary
- `seen()` now removes the `hidden` CSS class in addition to clearing inline `style.display`
- `toggle()` (show branch) applies the same fix for consistency

## Context
Elements initially hidden via the `hidden` CSS class were not revealed by `seen()` since it only cleared inline styles. The fix relies on `classList.remove()` being a no-op when the class is absent, and on `style.display = ''` deferring to CSS cascade so elements with other display-related classes (e.g. `d-flex`) are correctly restored.

### DIFF
--- a/core/dom/dom.ui.js
+++ b/core/dom/dom.ui.js
@@ -58,7 +58,7 @@ Element.prototype.isHidden = function() {
   return (this.offsetParent === null)
 }
 Element.prototype.seen = function() {
-  this.classList.remove('hidden')
+  this.removeClass('hidden')
   this.style.display = ''
   return this
 }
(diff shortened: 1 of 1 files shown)


Mips2648 (approved):

Salvialf:
`unseen()` now adds the `hidden` CSS class instead of setting `style="display:none"` inline.
`seen()` and `toggle()` already handled this class, this makes all three consistent.

Elements hidden via `unseen()` will no longer have an inline style in the DOM.
`seen()` clears both class and inline style for backward compatibility with existing HTML using `style="display:none"`.

Mips2648 (approved):

zoic21 (approved):

PR #3352, opened 2026-05-18
Fix jeedom update via api

Mips2648:
## Description
- fix jeedom;:update() calls
- refacto 

### Suggested changelog entry
- Correction des actions d'update via API

Fixes #

## Types of changes

## PR checklist

### DIFF
--- a/core/class/jeedom.class.php
+++ b/core/class/jeedom.class.php
@@ -880,10 +880,8 @@ class jeedom {
 	public static function update(array $_options = []) {
 		log::clear('update');
 		$params = '';
-		if (count($_options) > 0) {
-			foreach ($_options as $key => $value) {
-				$params .= '"' . $key . '"="' . $value . '" ';
-			}
+		foreach ($_options as $key => $value) {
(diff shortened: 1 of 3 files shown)


core/class/jeedom.class.php
```diff
@@ -880,10 +880,8 @@ public static function restore(string $_backup = '', bool $_background = false)
 	public static function update(array $_options = []) {
 		log::clear('update');
 		$params = '';
-		if (count($_options) > 0) {
```
zoic21, line 883: If we remove count it's fire a warning alert in log when we do a foreach on empty or null array

Mips2648: Never saw that.
- anyway the `$_options` array cannot be null here, it is impossible
- a foreach on an empty array is perfectly fine and don't create any warning, it simply does nothing

It is quite the opposite actually, if `$_options` would be null (which cannot happen anyway), doing a count() on a null will throw an error similar to: `count() expecting array, object or countable`

zoic21 (approved):

Salvialf (approved):

PR #3401, opened 2026-06-27
Add history retention mode

Salvialf:
## Summary

- Add a new `historyRetentionMode` configuration per **numeric info command** (the only subtype for which smoothing options are available)
- When active, forces `historizeMode` to `none` and `history::smooth` to disabled to prevent conflicts with real-time smoothing
- The corresponding fields are hidden in the UI when a retention mode is selected

image

## Retention options

| Option | Use case |
|--------|----------|
| **Last value of the day** | Commands updated frequently where only the end-of-day value matters: daily energy index, gas or water meter, daily rainfall accumulation, real-time price or tariff |
| **First and last value of the day** | Cumulative counters where the daily delta matters (last - first): electricity, gas or water index to calculate daily consumption |
| **Last value of the month** | Commands representing a consolidated monthly state: peak power demand, monthly usage total, subscription consumption counter |
| **First and last value of the month** | Same as above but where the monthly delta is needed, typically for billing-period index readings |

## Notes

- Retention is applied in real time on each call to `addHistoryValue`, before the value is saved
- When retention is set back to none, `historizeMode` and `history::smooth` are restored to their last saved values if they were not previously overridden by a retention save. Otherwise they reset to their defaults (`avg` and empty)
- Documentation update required in `docs/fr_FR/display.md`: add `historyRetentionMode` field description under the history section of the info command configuration tab

### DIFF
--- a/core/class/cmd.class.php
+++ b/core/class/cmd.class.php
@@ -2762,6 +2762,36 @@ class cmd {
 			return;
 		}
 
+		$retentionMode = $this->getConfiguration('historyRetentionMode', 'none');
+		if ($retentionMode !== 'none') {
+			$ts = ($_datetime !== '') ? strtotime($_datetime) : time();
+			$dayEnd = date('Y-m-d 23:59:59', $ts);
+
+			switch ($retentionMode) {
+				case 'current_day':
+					history::removes($this->getId(), date('Y-m-d 00:00:00', $ts), $dayEnd);
+					break;
+
+				case 'current_day_keep_first':
+					$histories = $this->getHistory(date('Y-m-d 00:00:00', $ts), $dayEnd);
+					if (count($histories) > 1) {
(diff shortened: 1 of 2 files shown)


MrWaloo:
My 2 cents: instead of "Aucun" for retention mode, the text should be "Toutes les valeurs" or something like this.
Setting the retention mode to "None" (aucun) could make the user think, there is no retention at all... Or it should be called "retention **limitation** mode", then "None" would make sense.

zoic21 (approved):

Mips2648 (approved):

Salvialf:
Hi @MrWaloo,

Thanks for raising this, "Aucun" was indeed ambiguous in this context. Changed to "Normal" to clearly convey that the default behavior applies (no limitation). The field label "Mode de rétention" stays as is since it accurately describes the feature when a non-default option is selected.

PR #3421, opened 2026-07-14
Avoid config pollution when a plugin lookup fails on a non-existent id

Salvialf:
## Context

Reported on the [Jeedom community forum](https://community.jeedom.com/t/plugin-logmanager-fichier-logmanager-class-php-introuvable-jeedom-4-6-x/149955/25): `plugin::byId()` calls `forceDisablePlugin()` on every failed lookup, even when the id was never a real plugin (e.g. a caller using `byId()` combined with try/catch as an "is this plugin installed" check, which is a legitimate pattern). `forceDisablePlugin()` unconditionally wrote an `active=0` config row for that id, a write side effect in what amounts to a read/existence check.

## Fix

`plugin::forceDisablePlugin()` now only writes the `active=0` row when the id was already marked active. This removes the write side effect for ids that were never real plugins, while still disabling a plugin that was genuinely active and became unresolvable.

### DIFF
--- a/core/class/plugin.class.php
+++ b/core/class/plugin.class.php
@@ -179,7 +179,9 @@ class plugin {
 	}
 
 	public static function forceDisablePlugin($_id) {
-		config::save('active', 0, $_id);
+		if (config::byKey('active', $_id) == 1) {
+			config::save('active', 0, $_id);
+		}
 		$values = array(
 			'eqType_name' => $_id,
 		);
--- a/core/class/plugin.class.php
+++ b/core/class/plugin.class.php
@@ -179,7 +179,7 @@ class plugin {
 	}
 
 	public static function forceDisablePlugin($_id) {
-		if (config::byKey('active', $_id) == 1) {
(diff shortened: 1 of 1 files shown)


zoic21 (approved):

core/class/plugin.class.php
```diff
@@ -179,7 +179,9 @@ public static function byId($_id, $_full = false) {
 	}
 
 	public static function forceDisablePlugin($_id) {
-		config::save('active', 0, $_id);
+		if (config::byKey('active', $_id) == 1) {
```
Mips2648: I suggest to write: config::byKey('active', $_id, 0) to force default value to 0 and make it more explicit, what do you think?

Salvialf: It's strictly equivalent, `'' == 1` is false, I'm only matching values equal to 1. I can add it for explicitness if you prefer, but the default `''` already behaves the same.

Edit: added 0 as default

Mips2648 (approved):

PR #3434, opened 2026-07-21
Fix transformations and pie chart handling on real-time history graph updates

zoic21:
## Description

Real-time history graph updates were adding raw command values directly to the chart, without applying the transformations configured when the series was created.

As a result, series using the `calcul` callback could display correctly converted historical data but receive unconverted real-time points. For example, a solar production value of `445 W` was added as `445 kWh` instead of `0.445 kWh`, causing the chart scale to jump unexpectedly.

This change:

* Stores the `calcul` callback in the chart command configuration.
* Applies `calcul` to real-time values before adding them to the series.
* Applies `invertData` consistently to real-time updates.
* Keeps real-time points consistent with the initially loaded historical data.

## Testing

Tested with the Energy3 plugin using a solar production command reported in watts and converted to kilowatt-hours for display. Historical and real-time points now use the same scale, and the chart axis no longer jumps when a new command value is received.

---
**Edit by @Salvialf:**

Pie charts had a separate issue: since their single series carries the `id` of only the first added command, any real-time update for that command matched the series and appended an unnamed point to it instead of updating the corresponding slice, causing pie charts to accumulate ghost slices over time.

Additional changes:

- Pie chart slices are now updated in place on real-time updates instead of accumulating new points, matching each slice by command id.
- `calcul` is applied to the initial pie slice value too, so real-time updates don't cause a value jump on the first update. `invertData` is not applied to pie charts, since a negative slice value has no valid meaning for a pie's percentage-based rendering.
- Aligned the `calcul` guard used on initial load with the `typeof === 'function'` check used on real-time updates.
- Restored braces on single-line conditions in `graphUpdate` to match the rest of the file.

### DIFF
--- a/core/js/history.class.js
+++ b/core/js/history.class.js
@@ -156,13 +156,15 @@ jeedom.history.generatePlotBand = function(_startTime, _endTime) {
 
 jeedom.history.graphUpdate = function(_params) {
   for (const i in _params) {
-    if(_params[i].cmd_id == ''){
-      continue;
-    }
-    for(const chart in jeedom.history.chart){
-      for(const serie in jeedom.history.chart[chart].chart.series){
-        if(jeedom.history.chart[chart].chart.series[serie] && jeedom.history.chart[chart].chart.series[serie].options.id == _params[i].cmd_id){
-          jeedom.history.chart[chart].chart.series[serie].addPoint([Date.now()+(-1*(new Date()).getTimezoneOffset()*60*1000),_params[i].value])
+    if (_params[i].cmd_id == '') continue
+    for (const chart in jeedom.history.chart) {
+      const cmd = jeedom.history.chart[chart].cmd[_params[i].cmd_id]
+      for (const serie in jeedom.history.chart[chart].chart.series) {
+        if (jeedom.history.chart[chart].chart.series[serie]?.options.id == _params[i].cmd_id) {
+          let value = _params[i].value
+          if (typeof cmd?.calcul === 'function') value = cmd.calcul(value)
+          if (cmd?.option?.invertData) value = -value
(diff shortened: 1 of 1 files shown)


Salvialf (commented):
Thanks for the fix, transformations on real-time points were indeed missing. A few points:

**Pie charts (community topic [149797](https://community.jeedom.com/t/jeedom-4-6-bogue-mise-a-jour-graphique-charts-type-camembert-dans-design/149797))**

This doesn't fix the reported pie chart issue, which comes from the same function: for pies the series is created with `id: _params.cmd_id` (first cmd), so `graphUpdate` matches it and `addPoint([x, y])` appends a new unnamed slice ("Slice") on every real-time value. Slices accumulate until the browser freezes. Also note `chart[el].cmd[]` is never populated for pies (the assignment at L1098 is inside the non-pie branch), so `cmd` will always be `undefined` there and the new transformations won't apply anyway.

Minimal fix, consistent with the existing guards at L1148/L1322/L1431/L1488:

```js
for (const chart in jeedom.history.chart) {
  if (jeedom.history.chart[chart].type == 'pie') {
    continue
  }
  ...
```

Properly updating the existing slice (`point.update()`) would need a cmd_id to point mapping for pies, probably out of scope here.

**Guard consistency**

Initial load uses `if (_params.calcul)` (L738) while this uses `typeof cmd?.calcul === 'function'`. The typeof check is safer, might be worth aligning L738 on it.

**Style nit**

Minor preference: I'd keep braces on the single-line conditions (`cmd_id` guard, `calcul`, `invertData`) as in the original code. One-liners are easy to misread and error-prone when a statement gets added later.

```js
if (_params[i].cmd_id == '') {
  continue
}
...
if (typeof cmd?.calcul === 'function') {
  value = cmd.calcul(value)
}
if (cmd?.option?.invertData) {
  value = -value
}
```

Salvialf (approved):

Mips2648 (approved):

PR #3477, opened 2026-08-27
Fix icon selector search crashing and not filtering object background images

Salvialf:
`document.querySelector('#md_iconSelector div.tab-pane.active')` is used to detect which tab is currently shown, but `#tabicon` and `#tabobjectbg` both had the `active` class hardcoded in the PHP template, with nothing ever toggling it between them. `querySelector` always returned `#tabicon` first, regardless of which pane was actually visible.

Two separate issues came from this and from an unrelated array bounds bug, both introduced in #2241:

- Typing in the search box while the icon tab wasn't showing (or in some cases even while it was) crashed with `Cannot read properties of null (reading 'parentNode')` in `iconTreeOnScroll()`: it always looked up `#tabicon`'s legends against the icon tree regardless of which tab was active, and separately, `legends[i - 1]` could be `undefined` depending on scroll position, unrelated to the active tab.
image

- Searching within the object background image gallery never filtered anything, since the search handler scoped its filtering to whatever `.tab-pane.active` resolved to (always `#tabicon`) instead of `#tabobjectbg`.

Fix:

- Guard `iconTreeOnScroll()` to only run when the icon tab is actually active, and make its DOM lookups null-safe.
- Make `active` mutually exclusive between `#tabicon` and `#tabobjectbg`, based on `$objectId`, so tab-active detection reflects what's actually shown.

Noticed while working on the Font Awesome 7.3.1 upgrade (separate branch/PR), unrelated to it.

### DIFF
--- a/desktop/modal/icon.selector.php
+++ b/desktop/modal/icon.selector.php
@@ -264,21 +264,24 @@ include_file('3rdparty', 'tree/tree', 'js');
         this.icon_tree.reload()
       },
       iconTreeOnScroll: function() {
-        var view = document.querySelector('#md_iconSelector .tab-content').getBoundingClientRect();
-        var legends = document.querySelectorAll('#tabicon .imgContainer legend');
-        var i = 0;
+        if (document.querySelector('#md_iconSelector div.tab-pane.active')?.id != 'tabicon') {
+          return;
+        }
+        const view = document.querySelector('#md_iconSelector .tab-content').getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


Mips2648 (approved):

Mips2648:
Could close this: https://community.jeedom.com/t/erreur-js-qd-on-selectionne-une-icone/150737/2

zoic21 (approved):

PR #3487, opened 2026-09-03
Fix hardcoded result property and incorrect options variable

Mips2648:
## Description
Code wrongly assume that if there is only one "promptAttr", it is called "result" because that's the name given by default in this same file.

But if caller provide a custom input, and only one, if could be anything and if so, the return is undefined

plus fix of correct usage of options variable in date and time input pattern after review

### Suggested changelog entry
[DEV] - fix jeeDialog.prompt() to accept any name as "data-l1key"

Fixes #

## Types of changes

### DIFF
--- a/core/dom/dom.ui.js
+++ b/core/dom/dom.ui.js
@@ -1311,7 +1311,7 @@ var jeeDialog = (function () {
               if (typeof _callback === 'function') {
                 var data = event.target.closest('div.jeeDialog').querySelector('div.jeeDialogContent').getJeeValues('.promptAttr')[0]
                 var key = event.target.closest('button').getAttribute('data-type')
-                if (Object.keys(data).length == 1) data = data.result
+                if (Object.keys(data).length == 1) data = Object.values(data)[0]
                 if (data == '') data = null
                 _callback.apply(this, [data, key])
               }
--- a/core/dom/dom.ui.js
+++ b/core/dom/dom.ui.js
@@ -19,7 +19,7 @@
 /* jeeDOM UI functionnalities
 */
 
-domUtils.showLoading = function(_timeout) {
+domUtils.showLoading = function (_timeout) {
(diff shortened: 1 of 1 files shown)


Mips2648:
To provide a bit more context, here is an example:
If we wrote this code with `data-l1key="newUserPswd"` the value of `myValue` in the callback is undefined:
```JS
let content = '<input class="promptAttr" data-l1key="newUserPswd" autocomplete="off" type="password" placeholder="{{Mot de passe}}">'
    jeeDialog.prompt({
      title: "{{Quel est le nouveau mot de passe ?}}",
      message: content,
      inputType: false,
      callback: function (myValue) {
        if (myValue) {
        }
...
```

but if in the same code we use `data-l1key="result"` then `myValue` contains the expected value from the input. 
```JS
let content = '<input class="promptAttr" data-l1key="result" autocomplete="off" type="password" placeholder="{{Mot de passe}}">'
    jeeDialog.prompt({
      title: "{{Quel est le nouveau mot de passe ?}}",
      message: content,
      inputType: false,
      callback: function (myValue) {
        if (myValue) {
        }
...
```

This only happen if there is only one input, example below works already as expected:
```JS
    let content = '<input class="promptAttr" data-l1key="newUserLogin" autocomplete="off" type="text" placeholder="{{Identifiant}}">'
    content += '<input class="promptAttr" data-l1key="newUserMdp" autocomplete="off" type="password" placeholder="{{Mot de passe}}">'
    jeeDialog.prompt({
      title: "{{Ajouter un utilisateur}}",
      message: content,
      inputType: false,
      callback: function (result) {
        if (result) {
          const user = [{
            login: result.newUserLogin,
            password: result.newUserMdp
          }]
...
```

Salvialf (approved):

zoic21 (approved):

Issue #3361, opened 2026-06-02
Move Python installation to a dedicated script


Salvialf:
As part of the Debian 13 compatibility effort (#3149), the installation process for Python needs to be harmonized with the rest of the setup.

Currently, Python is installed directly within the main core installation script, even though it is not required for the core itself (only for plugins). Other dependencies that are installed by default, such as Node.js and Composer, are handled via dedicated scripts to maintain a clean separation of concerns.

**Proposal:**
Move the Python installation logic to a dedicated script `resources/install_python.sh`, aligning it with the existing pattern used for other default dependencies.

**Rationale:**
- Harmonizes the installation method: Python will be treated like Node.js and Composer (default installation via a separate script).
- Keeps the main core installation script clean and focused on core-specific logic.

PR #3491, opened 2026-09-04
Rework health() diagnostics and fix inconsistent checks

Salvialf:
Fixes several real bugs found while reviewing `jeedom::health()` line by line, and reworks the whole method for consistency and accuracy.

## Before / After
image
image

## Bug fixes (present before this change)
- Apache PrivateTmp check compared an untrimmed `shell_exec(... | wc -l)` result directly as a boolean; `"0\n"`/`"1\n"` are both non-empty, non-`"0"` strings in PHP (therefore truthy), so the row always showed OK regardless of whether PrivateTmp was actually enabled.
- Same untrimmed-`wc -l` issue independently broke the "sufficient memory" and "I/O errors" (second counter) checks, and the "system load" check's `nproc` comparison, all of which compared the raw string to a number.
- "System load" compared the load average to a hardcoded `20`, regardless of the actual number of CPU cores, so the same threshold applied identically to a 2-core Raspberry Pi and a 32-core server. It now compares against `nproc`.
- The Debian version check only verified the floor (`os::min`), never the ceiling (`os::max`); a system newer than the supported range was never flagged.
- PHP version floor was hardcoded to `5.5` with a comment about "version 5.4.x" that predates this codebase's actual PHP support policy by a decade; it's now `7.4` (`composer.json`'s declared floor), with a warning tier for `7.4`-`8` now that Debian 12/PHP 8.2 is the OS floor.
- I/O error check only ever looked for `mmcblk0`, missing CRC errors on hardware where storage enumerates as `mmcblk1` (affects some official boxes).

## Reworked checks (behavior intentionally changed, not simple bugs)
- Kernel version is no longer a second Debian-version check. It now uses `php_uname()` instead of a `shell_exec('uname -a')` subprocess, and checks the actual kernel release against architecture-specific floors: `4.4` on x86_64/aarch64, and `armv7l` always warns given the number of existing 32-bit installs.
- Scenario check no longer calls `scenario::all()` (which hydrates every scenario as a full object) just to check if any exist; it's a lightweight `COUNT(*)`, only run when the engine is disabled.
- Swappiness never hard-fails anymore (always `2`/warning at worst): the check can't tell a native install (where `/etc/sysctl.conf` is genuinely fixable) from an LXC/container (where `vm.swappiness` is a host-level setting the user can't touch).
- Available swap dropped the old combined `(RAM + swap) < 1900Mo` absolute rule in favor of a plain free-percentage check (matching "Available memory"'s `15`/`20` tiers), after checking it against real hardware configurations; "no swap at all" is now `danger` instead of `warning`.
- Database size now goes through `cmd::autoValueArray(..., 'io')` like the rest of the page instead of the separate `sizeFormat()` helper, for a consistent Kio/Mio/Gio labeling instead of English KB/MB/GB.
- Python version check consolidates the old two always-`state: true` rows (Python 2 and Python 3) into one row that's only "OK" when Python 3 is actually present.

## New checks (nothing existed here before)
- "Mises à jour" and "Version core" now distinguish a pending **core** update (danger) from pending plugin/market updates (warning); previously always a flat OK/count with no severity distinction.
- Node.js version now has a real floor, read dynamically from `resources/install_nodejs.sh`'s `minVer` instead of just displaying whatever's installed with no check at all.
- Database connections now warn/fail as usage approaches `max_connections` (70%/90%); previously always `state: true`, no check.
- Network external now distinguishes "configured but failing" (danger) from "not configured" (warning) via `market::allowDNS`/`externalAddr`; previously a flat true/false with no such distinction.
- System date now has a distinct state for "cache never recorded yet" (fresh install/reset) instead of showing the same OK green a real recorded date would.
- Available memory/swap/disk space now show an absolute free/total breakdown (via the existing `'io'` cmd unit, 1024-based) alongside the percentage, not just the percentage.

## Other
- `OK`/`NOK` are literal, untranslated strings throughout (translations were breaking the intended universal status code).
- Every row now uses a section-specific variable name instead of a handful of blocks reusing `$value`/`$state`, adopted specifically to prevent stale-value bugs of the kind this rework itself introduced and caught along the way.
- Removed a dead code path (`update::listRepo()` health loop) that no repo class has ever implemented.

Python version check intentionally left without a version floor for now (same treatment as Node.js, a core-managed install script with a `minVer`, is planned separately, see jeedom/core#3361).

### DIFF
--- a/core/class/jeedom.class.php
+++ b/core/class/jeedom.class.php
@@ -249,7 +249,7 @@ class jeedom {
 		if (!$scenarioState) {
 			$nbScenario = DB::Prepare('SELECT COUNT(*) as nb FROM scenario', array(), DB::FETCH_TYPE_ROW)['nb'];
 			if ($nbScenario > 0) {
-				$scenarioComment = __('Activer le moteur de scénarios (Outils -> Scénarios)', __FILE__);
+				$scenarioComment = __('Activer le moteur de scénarios (Outils > Scénarios)', __FILE__);
 			} else {
 				$scenarioState = 2;
 				$scenarioComment = __('Le moteur de scénarios est désactivé sans scénarios existants', __FILE__);
--- a/core/class/jeedom.class.php
+++ b/core/class/jeedom.class.php
@@ -173,355 +173,432 @@ class jeedom {
 
 	public static function health() {
 		$return = array();
+		// Hardware
 		$return[] = array(
 			'name' => __('Matériel', __FILE__),
 			'state' => true,
-			'result' => jeedom::getHardwareName(),
-			'comment' => '',
+			'result' => ucfirst(jeedom::getHardwareName()),
+			'comment' => __('Type de matériel utilisé', __FILE__),
 			'key' => 'hardware'
 		);
 
+		// Updates
 		$nbNeedUpdate = update::nbNeedUpdate();
-		$state = ($nbNeedUpdate == 0) ? true : false;
+		$updateState = true;
+		if ($nbNeedUpdate > 0) {
+			$coreUpdate = update::byTypeAndLogicalId('core', 'jeedom');
+			$coreNeedUpdate = is_object($coreUpdate) && $coreUpdate->getStatus() == 'update';
+			if ($coreNeedUpdate) {
+				$updateState = false;
+			} else {
+				$updateState = 2;
+			}
+		}
 		$return[] = array(
-			'name' => __('Système à jour', __FILE__),
-			'state' => $state,
-			'result' => ($state) ? __('OK', __FILE__) : $nbNeedUpdate,
-			'comment' => '',
+			'name' => __('Mises à jour', __FILE__),
+			'state' => $updateState,
+			'result' => ($updateState === true) ? 'OK' : $nbNeedUpdate,
+			'comment' => ($updateState === true) ? __('Aucune mise à jour en attente', __FILE__) : __('Nombre de mises à jour en attente', __FILE__),
 			'key' => 'uptodate'
 		);
 
-		$state = (config::byKey('enableCron', 'core', 1, true) != 0) ? true : false;
+		// Start date
+		$product_name = config::byKey('product_name');
+		$isStarted = self::isStarted();
 		$return[] = array(
-			'name' => __('Cron actif', __FILE__),
-			'state' => $state,
-			'result' => ($state) ? __('OK', __FILE__) : __('NOK', __FILE__),
-			'comment' => ($state) ? '' : __('Erreur cron : les crons sont désactivés. Allez dans Réglages -> Système -> Moteur de tâches pour les réactiver', __FILE__),
-			'key' => 'cron::enable'
-		);
-
-		$state = (config::byKey('enableScenario') == 0 && count(scenario::all()) > 0) ? false : true;
-		$return[] = array(
-			'name' => __('Scénario actif', __FILE__),
-			'state' => $state,
-			'result' => ($state) ? __('OK', __FILE__) : __('NOK', __FILE__),
-			'comment' => ($state) ? '' : __('Erreur scénario : tous les scénarios sont désactivés. Allez dans Outils -> Scénarios pour les réactiver', __FILE__),
-			'key' => 'scenario::enable'
-		);
-
-		$state = self::isStarted();
-		$return[] = array(
-			'name' => __('Démarré', __FILE__),
-			'state' => $state,
-			'result' => ($state) ? __('OK', __FILE__) . ' ' . file_get_contents(self::getTmpFolder() . '/started') : __('NOK', __FILE__),
-			'comment' => '',
+			'name' => __('Démarrage', __FILE__),
+			'state' => $isStarted,
+			'result' => ($isStarted) ? file_get_contents(self::getTmpFolder() . '/started') : 'NOK',
+			'comment' => ($isStarted) ? __('Date de démarrage de', __FILE__) . ' ' . $product_name : $product_name . ' ' . __('est en cours de démarrage, patienter quelques minutes puis rafraîchir la page', __FILE__),
 			'key' => 'isStarted'
 		);
 
-		$state = self::isDateOk();
-		$cache = cache::byKey('hour');
-		$lastKnowDate = $cache->getValue();
-		if ($lastKnowDate === "") {
-			$lastKnowDate = 0;
+		// Crons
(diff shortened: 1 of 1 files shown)


Mips2648 (approved):

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
 		);
 
-		$state = (config::byKey('enableScenario') == 0 && count(scenario::all()) > 0) ? false : true;
+		// System date
+		$dateState = self::isDateOk();
+		$lastKnowDate = cache::byKey('hour')->getValue(0);
+		if ($lastKnowDate == 0) {
+			$dateState = 2;
+			$lastKnowDate = '-';
+			$dateComment = __("Aucune heure enregistrée, patienter 1 heure maximum", __FILE__);
```
Salvialf:
```suggestion
			$dateComment = __("Aucune heure enregistrée pour le moment, patienter jusqu'à la prochaine heure", __FILE__);
```

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
-
 		$return[] = array(
-			'name' => __('Version OS', __FILE__),
-			'state' => (system::getDistrib() == 'debian' && version_compare(system::getOsVersion(), config::byKey('os::min'), '>=')),
-			'result' => system::getDistrib() . ' ' . system::getOsVersion(),
-			'comment' => '',
+			'name' => __('Version système', __FILE__),
+			'state' => $osState,
+			'result' => ucfirst($distrib) . ' ' . $osVersion,
+			'comment' => ($osState) ? __("Version du système", __FILE__) : __("Ce système n'est pas officiellement supporté (voir la documentation sur la compatibilité logicielle)", __FILE__),
```
Salvialf:
```suggestion
			'comment' => ($osState) ? __('Version du système', __FILE__) : __("Ce système n'est pas officiellement supporté (voir la documentation sur la compatibilité logicielle)", __FILE__),
```

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
 		$return[] = array(
-			'name' => __('Version PHP', __FILE__),
-			'state' => $state,
-			'result' => phpversion(),
-			'comment' => ($state) ? '' : __('Si vous êtes en version 5.4.x on vous indiquera quand la version 5.5 sera obligatoire', __FILE__),
-			'key' => 'php::version'
+			'name' => __('Persistance du cache', __FILE__),
+			'state' => $cacheState,
+			'result' => ($cacheState) ? 'OK' : 'NOK',
+			'comment' => ($cacheState) ? __('Persistance du cache après redémarrage', __FILE__) : __('Certaines informations peuvent être perdues après redémarrage, attendre 30 minutes maximum ou démarrer la tâche cache::persist manuellement (Réglages > Système > Moteur de tâches)', __FILE__),
```
Salvialf:
```suggestion
			'comment' => ($cacheState) ? __('Persistance du cache après redémarrage', __FILE__) : __("Certaines informations peuvent être perdues après redémarrage, patienter jusqu'à la prochaine demi-heure ou démarrer la tâche cache::persist manuellement (Réglages > Système > Moteur de tâches)", __FILE__),
```

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
 		$return[] = array(
-			'name' => __('Configuration réseau interne', __FILE__),
-			'state' => $state,
-			'result' => ($state) ? __('OK', __FILE__) : __('NOK', __FILE__),
-			'comment' => ($state) ? '' : __('Allez sur Réglages -> Système -> Configuration -> onglet Réseaux, puis configurez correctement la partie réseau', __FILE__),
-			'key' => 'network::internal'
+			'name' => __('Mémoire disponible', __FILE__),
+			'state' => $memState,
+			'result' => $percentMem . '% (' . $availableMem[0] . $availableMem[1] . '/' . $totalMem[0] . $totalMem[1] . ')',
+			'comment' => __("Pourcentage de mémoire disponible (libre/total)", __FILE__),
```
Salvialf:
```suggestion
			'comment' => __('Pourcentage de mémoire disponible (libre/total)', __FILE__),
```

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
 		$return[] = array(
-			'name' => __('Node', __FILE__),
-			'state' => true,
-			'result' => $value,
-			'comment' => '',
-			'key' => 'node::version'
+			'name' => __('Swap disponible', __FILE__),
+			'state' => $swapState,
+			'result' => $swapResult,
+			'comment' => __("Pourcentage de swap disponible (libre/total)", __FILE__),
```
Salvialf:
```suggestion
			'comment' => __('Pourcentage de swap disponible (libre/total)', __FILE__),
```

core/class/jeedom.class.php
```diff
@@ -173,355 +173,450 @@ public static function deadCmd() {
...
+			'result' => $nb_active_connection['Value'] . '/' . $max_used_connection['Value'] . '/' . $allow_connection['Value'],
+			'comment' => ($dbConnState === true) ? __('Nombre de connexions à la base de données (actives/maximales/autorisées)', __FILE__) : __('Le nombre maximal de connexions à la base de données approche la limite autorisée (actives/maximales/autorisées)', __FILE__),
+			'key' => 'database::connexion'
 		);
 
-		$value = shell_exec('node --version');
+		// Available swap
+		$swapState = false;
+		$swapResult = 'NOK';
+		if ($systemMemInfo['SwapTotal'] > 0) {
```
Mips2648, line 551: There are valid use cases where there is no swap on the system; e.g. proxmox LXC with zfs storage don't have swap because it is not supported (it's my case ;-)); but could be anyone having configured to have no swap for whatever reason
Previously it was only a warning, which was already disturbing, a NOK is definitively even more annoying; we risk some questions around this topic.

Maybe we should only display the values (free / total) which are facts
and change the status (which is an interpretation) only based on the ratio? This way having 0/0 would be "ok" but having 0/xxx would not

Salvialf: No swap at all is now treated as OK rather than NOK (and the same reasoning applies to the swappiness check, which is now skipped entirely when there's no swap). Fixed in cfa9ca9.

zoic21 (approved):

PR #3512, opened 2026-09-11
Update Node.js version to 24 and adjust armv6 handling

NebzHB:
Updated Node.js version to 24 for installation and minimum acceptance. Adjusted handling for unsupported architectures and provided a fallback to version 22 for armv6.

this probably breaks support for armv6 (Jeedom Mini, RPI 1, RPI Zero)

### DIFF
--- a/resources/install_nodejs.sh
+++ b/resources/install_nodejs.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-installVer='22' 	#NodeJS major version to be installed
-minVer='22'	      #min NodeJS major version to be accepted
+installVer='24' 	#NodeJS major version to be installed
+minVer='24'	      #min NodeJS major version to be accepted
 
 # vérifier si toujours nécessaire, cette source traine encore sur certaines smart et si une source est invalide -> nodejs ne s'installera pas
 if ls /etc/apt/sources.list.d/deb-multimedia.list* &>/dev/null; then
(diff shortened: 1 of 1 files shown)


resources/install_nodejs.sh
```diff
@@ -84,6 +84,18 @@ then
...
+then
+  today=$(date +%Y%m%d)
+  if [[ "$today" > "20260831" ]];
+  then
+  echo "== ATTENTION Debian 11 Bullseye n'est officiellement plus supportée depuis le 31 aout 2026, merci de mettre à jour votre distribution !!!"
+  exit 1
+fi
+fi
+
 #x86 32 bits not supported by nodesource anymore
```
Salvialf, line 99:
```suggestion
#version to install for armv6 (to check on https://unofficial-builds.nodejs.org), nodejs 24 not supported anymore, installing nodejs 22 on best effort
if [[ $arch == "armv6l" ]]
then
    minVer='22'
fi

#x86 32 bits not supported by nodesource anymore
```

`minVer` needs to reflect what's actually achievable per architecture, and it needs to be set here, before the first version check just below. That check (and the post-install one further down) both compare against `minVer`. Without this, on `armv6l`, `minVer` stays at the global `24`: the first check will never pass after installing `22.23.2` (the only option there), so every future run will think Node needs (re)installing and go through the whole install again, indefinitely. Setting it here fixes both checks in one place.

resources/install_nodejs.sh
```diff
@@ -118,18 +130,10 @@ else
...
-      armVer="14.21.3"
-    elif [[ $installVer == "16" ]]; then
-      armVer="16.20.2"
-    elif [[ $installVer == "18" ]]; then
-      armVer="18.18.0"
-    elif [[ $installVer == "20" ]]; then
-      armVer="20.8.0"
-    fi
+    #version to install for armv6 (to check on https://unofficial-builds.nodejs.org), nodejs 24 not supported anymore, installing nodejs 22 on best effort
+    $installVer=22
```
Salvialf, line 134:
```suggestion
```

No longer needed here: `minVer` is already set for `armv6l` above. `installVer` isn't read again in this branch anyway (`armVer` is hardcoded right below), so `$installVer=22` (invalid syntax on top of that) can just be dropped.

Mips2648 (approved):

Salvialf (approved):

PR #3514, opened 2026-09-12
Chore: Remove the OS/package update feature from Jeedom.

Mips2648:
## Description

Remove the obsolete OS/package update feature: 
- Python 2 is no longer supported, 
- Python 3 package updates are becoming obsolete with Debian 12 and 13 (because they are managed in plugins venv),
- only APT updates remained available on DIY boxes, where users can and should manage them themselves so it does not make sense to keep it in the core.

I took the opportunity to also refactor the `var` declarations in update.js and replace them by `let` and `const`

### Suggested changelog entry

- Breaking change: Remove the OS/package update feature from Jeedom (it was only available on DIY boxes)

Fixes https://community.jeedom.com/t/core-develop-deb13-retours/150813

## Types of changes

- [X] Breaking change _(fix or feature that would cause existing functionality to change)_

### DIFF
--- a/desktop/js/update.js
+++ b/desktop/js/update.js
@@ -21,13 +21,11 @@ if (!jeeFrontEnd.update) {
     replaceLogLines: ['OK', '. OK', '.OK', 'OK .', 'OK.'],
     regExLogProgress: /\[PROGRESS\]\[(\d.*)]/gm,
     updtDataTable: null,
-    osDataTable: null,
     init: function () {
       window.jeeP = this
       this.hasUpdate = false
       this.progress = -2
       this.alertTimeout = null
-      this.osUpdateChecked = 0
       //___log interceptor beautifier___
       this.prevUpdateText = ''
       this.newLogClean = '<pre id="pre_updateInfo_clean" style="display:none;"><i>{{Aucune mise à jour en cours.}}</i></pre>'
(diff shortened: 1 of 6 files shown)


desktop/js/update.js
```diff
@@ -22,7 +22,7 @@ if (!jeeFrontEnd.update) {
     regExLogProgress: /\[PROGRESS\]\[(\d.*)]/gm,
     updtDataTable: null,
     osDataTable: null,
```
Salvialf:
```suggestion
```

Leftover from the removed OS/package tab no longer read anywhere.

desktop/js/update.js
Salvialf:
```diff
-      this.osUpdateChecked = 0
```
```suggestion
```
Same here `osUpdateChecked` is dead since printOsUpdate/addOsUpdate are gone.

desktop/js/update.js
Mips2648: good catch, thx, resolved with commit https://github.com/jeedom/core/pull/3514/changes/90c239f5ca689f040dc5593a2a4e206313df021e

desktop/js/update.js
Mips2648: good catch, thx, resolved with commit https://github.com/jeedom/core/pull/3514/changes/90c239f5ca689f040dc5593a2a4e206313df021e

Salvialf (approved):

PR #3533, opened 2026-09-19
Fix 2FA QR code display when imagick is not installed

limad:
Without the imagick extension, Google2FA::getQRCodeInline() returns raw SVG markup instead of a data URL. Injected in , the double quotes of the SVG close the attribute early: a broken image icon and a stray `" />` are displayed next to the QR code.

Wrap the SVG in a base64 data URL when the result is not already one.

## Types of changes

## PR checklist

- [ ] I have checked there is no other PR open for the same change.
- [ ] I have read the [[La ligne directrice pour contribuer à ce projet / Contribution guidelines for this project](.github/CONTRIBUTING.md)).
- [ ] I grant the project the right to include and distribute the code under the GNU.
- [ ] I have verified that the code complies with the projects coding standards.


Mips2648:
Thanks for the PR, this fix makes sense and we’ll accept it.
Before merging, we’ll also add two improvements:
- shipping Imagick by default so the inline PNG path is always available, 
- and HTML‑encoding the value before injecting it into the src attribute to prevent any attribute‑breaking cases like this one.

PR #3535, opened 2026-09-21
Fix apt/dpkg lock race during installation

Salvialf:
During install, `install/consistency.php` calls `system::checkAndInstall()` to check and fix missing system packages. When a package is missing, that fix runs in the background: it is scheduled via `at now`, which `atd` only picks up on the next minute boundary, not immediately.

That deferred job starts with `killall apt apt-get unattended-upgr` and forcibly removes the dpkg/apt lock files before reinstalling the missing package. Since `install.php` doesn't wait for this background job, the installer keeps going (e.g. to the nodejs install step), so this "fix" script can fire while another legitimate `apt`/`dpkg` operation is still running elsewhere in the installer, killing it mid-transaction and/or stripping its lock out from under it.

This explains the `Could not get lock /var/lib/dpkg/lock-frontend` reports and inconsistent nodejs installs seen on the forum:
- https://community.jeedom.com/t/appel-aux-testeurs-gros-lot-de-nouveautes-on-a-besoin-de-vos-retours/150759/2
- https://community.jeedom.com/t/installation-vm-en-debian-13-non-fonctionnelle-jeedom-toujours-en-cours-de-demarrage/150869/24

`python3-venv` made this worse: it's been required by `install/packages.json` since April 2024 but was never added to `install.sh`'s own apt install list, so this deferred fix path was triggered on literally every fresh install, not just when something failed earlier.

## Fix

- Run the package check in foreground during install, so any missing package is fixed synchronously and in order instead of being dispatched via `at`.
- Add `python3-venv` to `install.sh` so it no longer relies on that fallback path for a package that's needed on every install anyway.


zoic21 (approved):

install/install.sh
```diff
@@ -64,7 +64,7 @@ step_2_mainpackage() {
   apt-get update
   apt_install chrony ca-certificates unzip curl sudo cron
   apt-get -o Dpkg::Options::="--force-confdef" -y install plocate tar telnet wget logrotate dos2unix htop iotop vim iftop smbclient
-  apt-get -y install git python3 python3-pip
+  apt-get -y install git python3 python3-pip python3-venv
```
Mips2648, line 67: I think we can so we should remove all python related packages here
- they are already listed in packages.json used by various fallback mechanism; maintaining 2 distinct lists is a risk of discrepancy
- they will be installed explicitly and automatically by system::checkAndInstall() in case a plugin needs to install python dependencies
- and the only need to have these package is if a plugin with python dependencies is installed, before that core doesn't need them.

Mips2648: ok, let's keep it like this now, at least both list (this one and the one in packages.json) are in sync
potential cleaning will happen in another issue

Mips2648 (approved):

